PR #5, opened 2019-01-31
Delete CODE_OF_CONDUCT.md

SimonBoothroyd:
## Description
Removed the code of conduct document which was added as part of the MolSSI cookie cutter until it can be discussed further with the consortium.


jchodera:
I just want to note for the record that anything we do that deviates from best practices---including deleting a code of conduct or amending a code of conduct [already in use by hundreds of organizations](https://www.contributor-covenant.org/adopters)---should not be done lightly.

PR #9, opened 2019-02-01
Fix docs builds on RTD

SimonBoothroyd:
Moved any openforcefield imports from the global to the local scope to avoid having a docs dependance on openeye.


j-wags:
Waiting on Travis to accept.

dgasmith:
Cool, glad you pull this out and it looks the tools are working well!

PR #10, opened 2019-02-05
[WIP] Examples of different EstimatedQuantity class implentations.

SimonBoothroyd:
## Description
I have sketched out four possible variants of an EstimatedQuantity class which would store both a value and its uncertainty, and would propagate this uncertainty through simple arithmetic (+, -, scalar *, scalar /) operations.

The variants can be found in the [quantities module](https://github.com/openforcefield/propertyestimator/blob/EstimatedProperty/propertyestimator/utils/quantities.py) with some additional test code in [this test module](https://github.com/openforcefield/propertyestimator/blob/EstimatedProperty/propertyestimator/tests/test_quantities.py).

## Questions
Which design (if any) looks like the best way forward?

## Status
- [ ] Ready to go

### DIFF
--- a/propertyestimator/utils/quantities.py
+++ b/propertyestimator/utils/quantities.py
@@ -2,7 +2,6 @@
 their uncertainties
 """
 
-import math
 from collections import Sized
 from collections.abc import Iterable
 
@@ -10,11 +9,42 @@ import numpy as np
 from simtk import unit
 from uncertainties import ufloat
 
+from propertyestimator.utils import statistics
+
 
 class VanillaEstimatedQuantity:
     """A 'bare minimum' implementation of an `EstimatedQuantity` class, which employs
     the uncertainty package to linearly propagate uncertainties through simple arithmetic
     and scalar multiplication / division operations.
+
+    Examples
+    --------
+    To add two **independent** quantities together:
+
+    >>> from simtk import unit
+    >>>
+    >>> a_value = 5 * unit.angstrom
+    >>> a_uncertainty = 0.03 * unit.angstrom
+    >>>
+    >>> b_value = 10 * unit.angstrom
+    >>> b_uncertainty = 0.04 * unit.angstrom
+    >>>
+    >>> quantity_a = VanillaEstimatedQuantity(a_value, a_uncertainty)
+    >>> quantity_b = VanillaEstimatedQuantity(b_value, b_uncertainty)
+    >>>
+    >>> quantity_addition = quantity_a + quantity_b
+
+    To subtract one **independent** quantity from another:
+
+    >>> quantity_subtraction = quantity_b - quantity_a
+
+    To multiply by a scalar:
+
+    >>> quantity_scalar_multiply = quantity_a * 2.0
+
+    To divide by a scalar:
+
+    >>> quantity_scalar_divide = quantity_a / 2.0
     """
 
     @property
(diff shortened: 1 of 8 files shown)


propertyestimator/utils/quantities.py
```diff
@@ -0,0 +1,656 @@
...
+import numpy as np
+from simtk import unit
+from uncertainties import ufloat
+
+from propertyestimator.utils import statistics
+
+
+class VanillaEstimatedQuantity:
+    """A 'bare minimum' implementation of an `EstimatedQuantity` class, which employs
+    the uncertainty package to linearly propagate uncertainties through simple arithmetic
```
maxentile:
```suggestion
    the `uncertainties` package to linearly propagate uncertainties through simple arithmetic
```

j-wags:
This pull request **introduces 1 alert** and **fixes 3** when merging 89baac6c8084a1c2aa229410f07b653f0e3f1d82 into 813a04a903c2cd95009544e06c6858489612db12 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-81c14ed3f5e68a239c6e3aa6a313c7d1a2c9ac42)

**new alerts:**

* 1 for \_\_eq\_\_ not overridden when adding attributes

**fixed alerts:**

* 1 for Unnecessary pass
* 1 for Except block handles &#39;BaseException&#39;
* 1 for Unused import

---

propertyestimator/utils/quantities.py
```diff
@@ -0,0 +1,656 @@
...
+            raise DependantValuesException()
+
+        return super(TaggedEstimatedQuantity, self).__sub__(other)
+
+
+class ExplicitEstimatedQuantity:
+    """An implementation of an `EstimatedQuantity` class, which is almost identical
+    to the VanillaEstimatedQuantity implementation, except now all operators have been
+    replaced with explicit method calls.
+
```
maxentile: @SimonBoothroyd : Could you add a line or two about when we should use this vs `VanillaEstimatedQuantity`?

propertyestimator/utils/quantities.py
```diff
@@ -0,0 +1,656 @@
...
+    >>> quantity_a = TaggedEstimatedQuantity(a_value, a_uncertainty, 'calc_325262315:npt_production')
+    >>> quantity_b = TaggedEstimatedQuantity(b_value, b_uncertainty, 'calc_987234582:npt_production')
+    >>>
+    >>> quantity_addition = quantity_a + quantity_b
+
+    To subtract one **independent** quantity from another:
+
+    >>> quantity_subtraction = quantity_b - quantity_a
+
+    Attempting to add quantities from the same source (i.e probably correlated)
```
maxentile: Are there other checks that would make sense to consider here, in addition to checking for identical source tags?

propertyestimator/utils/quantities.py
```diff
@@ -0,0 +1,656 @@
...
+        result_value = result_ufloat.nominal_value * self_unit
+        result_uncertainty = result_ufloat.std_dev * self_unit
+
+        return VanillaEstimatedQuantity(result_value, result_uncertainty)
+
+    def __eq__(self, other):
+        if not isinstance(other, VanillaEstimatedQuantity):
+            return False
+
+        return self.value == other.value and self.uncertainty == other.uncertainty
```
andrrizzi: I guess this is a conservative implementation, but we probably need to be careful when defining equality in a statistical sense. This criterion will probably never be true unless the second quantity is a copy of the first. Using `np.isclose` may be unexpected behavior. I'm thinking that for equality checking, the safe option we be to add explicit methods rather than Python operators and raise an exception declaring that equating two statistical quantities is ambiguous (similar to `np.ndarray`) here.

maxentile: >I'm thinking that for equality checking, the safe option we be to add explicit methods rather than Python operators and raise an exception declaring that equating two statistical quantities is ambiguous (similar to `np.ndarray`) here.

I agree that is the safer (but less convenient) option here, maybe also for `>`, `<`, ... operators.

Note that the `uncertainties` package interprets comparison operators in a specific way, described in [section 4.3 of the tech guide](https://media.readthedocs.org/pdf/uncertainties-python-package/latest/uncertainties-python-package.pdf#page=26&zoom=100,0,96).

PR #11, opened 2019-02-05
Fix for openforcefield ForceField not being serializable

SimonBoothroyd:
## Description
I added two helper methods for serializing and deserializing an `openforcefield` `ForceField` class to / from a dict as the conda version of the class does not implement `__getstate__` or `__setstate__` methods. The rest of the codebase was then updated to use the new serialization methods, rather than the old and now missing get / set state ones.

The PR also includes some general clean-up as suggested by lgtm.

## Status
- [x] Ready to go


j-wags:
This pull request **introduces 2 alerts** and **fixes 4** when merging d81701744369b0e80444053750533b18077bc8bd into 813a04a903c2cd95009544e06c6858489612db12 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-53ce7236dfdeb5d689b0c99299fa1cef1aca6c5a)

**new alerts:**

* 1 for Unused import
* 1 for Suspicious unused loop iteration variable

**fixed alerts:**

* 1 for Unnecessary pass
* 1 for Except block handles &#39;BaseException&#39;
* 1 for Unused import
* 1 for Unreachable code

---

propertyestimator/utils/serialization.py
```diff
@@ -330,3 +331,40 @@ def deserialize_quantity(serialized):
             quantity_unit *= (getattr(unit, unit_name) ** power)
     quantity = unit.Quantity(serialized['unitless_value'], quantity_unit)
     return quantity
+
+
+def serialize_force_field(force_field):
+
```
j-wags: Could you add docstrings and comments? I'm not super up to date on the old FF implementation, so it took me a while to figure out what this is doing and how.

propertyestimator/utils/serialization.py
```diff
@@ -330,3 +331,40 @@ def deserialize_quantity(serialized):
...
+            string_value = file_buffer.getvalue().decode()
+            return_dictionary[index] = string_value
+
+            file_buffer.close()
+
+        return return_dictionary
+
+
+def deserialize_force_field(force_field_dictionary):
+
```
j-wags: Same w/ docstring and comments here

j-wags:
This pull request **fixes 4 alerts** when merging 62ef800f6bbd98fe7aa303dacd9892c0cde85ccb into 813a04a903c2cd95009544e06c6858489612db12 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-e9ba75610908a5e6b9992606fbb7d566d40605bf)

**fixed alerts:**

* 1 for Unnecessary pass
* 1 for Except block handles &#39;BaseException&#39;
* 1 for Unused import
* 1 for Unreachable code

---

setup.py
```diff
@@ -10,7 +10,7 @@
 try:
     with open("README.md", "r") as handle:
         long_description = handle.read()
-except:
+except IOError:
```
j-wags: Nice catch of specific error type.

propertyestimator/tests/test_utils_serialization.py
```diff
@@ -91,3 +92,13 @@ def test_polymorphic_dictionary():
...
+
+
+def test_force_field_serialization():
+
+    from openforcefield.typing.engines import smirnoff
+
+    force_field = smirnoff.ForceField(get_data_filename('forcefield/smirnoff99Frosst.offxml'))
+
+    serialized_force_field = serialize_force_field(force_field)
+    deserialized_force_field = deserialize_force_field(serialized_force_field)
```
j-wags: It would be good to test equality for some deserialized force field parameters here. Maybe check that `force_field` and `deserialized_force_field` have the same number of parameters, and that the SMIRKS+k value for the first bond parameter is the same in both?

propertyestimator/layers/simulation.py
```diff
@@ -683,7 +683,7 @@ def schedule_calculation(calculation_backend, storage_backend, layer_directory,
         force_field_path = path.join(layer_directory, 'force_field_{}'.format(data_model.parameter_set_id))
 
         with open(force_field_path, 'wb') as file:
-            pickle.dump(force_field, file)
+            pickle.dump(serialize_force_field(force_field), file)
```
j-wags: I'm not highly experienced with serialization, but this seems like it might be an unnecessary use of pickle. Is there anything wrong with `file.write(serialize_force_field(force_field))`? Pickle potentially introduces python-only behavior (https://www.datacamp.com/community/tutorials/pickle-python-tutorial#not)

Also, I generally try to avoid using `file` as a variable because Python seems to have it as a reserved keyword (I've never used the keyword, but the syntax highlighting indicates a potential problem). Maybe consider `file_obj` or `fo`.

SimonBoothroyd: I think the main reason is that the storage layer is written as an abstraction, which while the currently employed implementation stores objects on the disk, future implementations may store objects with a database for example. Using pickle gives this flexibility for moving away from local files 'easily'! But there could be a better way to handle this?

propertyestimator/tests/test_utils_serialization.py
SimonBoothroyd: Yeah I was trying to think if there are any good (read as 'simple') ways to do this that would give sufficient confidence that the two are actually the same!

propertyestimator/layers/simulation.py
SimonBoothroyd: Also I thought file was a builtin in Python 2 which got removed in 3? Could be wrong about this though!

propertyestimator/utils/serialization.py
SimonBoothroyd: Must've missed these - thanks for pointing them out!

propertyestimator/layers/simulation.py
j-wags: Ah, so `file` is a holdover from Python 2. I'm learning new things!

propertyestimator/tests/test_utils_serialization.py
j-wags: We do some counting tests in the `topology` branch. But I don't know how I'd write this in the old API. https://github.com/openforcefield/openforcefield/blob/topology/openforcefield/tests/test_forcefield.py#L158-L165

j-wags:
This pull request **fixes 4 alerts** when merging dfd95d2de99c8741830f38af61f1323b48042de5 into 813a04a903c2cd95009544e06c6858489612db12 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-2f0840ecd231d4d20d91d32d003472554fc20d65)

**fixed alerts:**

* 1 for Unnecessary pass
* 1 for Except block handles &#39;BaseException&#39;
* 1 for Unused import
* 1 for Unreachable code

---

PR #12, opened 2019-02-08
Implemented a workflow to estimate enthalpy of mixing

SimonBoothroyd:
## Description
The machinery was added to allow the estimation of more complex physical properties, such as enthalpy of mixing, which require parts of their workflow to be replicated.

This in part was accomplished by introducing a new `ProtocolReplicator` class which was added as an entry to the `WorkflowSchema`. To accommodate this, large parts of the `DirectCalculationGraph` and `ProtocolGroup` classes had to be rewritten to be more general, and offer better support for workflows which are not linear.

This PR also introduces the ability to pass lists of `ProtocolPath`'s as input to protocols, and have the paths be automatically swapped out for the values they point to at runtime.

## Status
- [x] Ready to go


propertyestimator/backends/dask.py
```diff
@@ -9,10 +9,6 @@
 from .backends import PropertyEstimatorBackend, PropertyEstimatorBackendResources
 
 
-# =============================================================================================
```
j-wags: Yeah, good call on removing these.

propertyestimator/layers/simulation.py
```diff
@@ -109,27 +164,161 @@ def __init__(self, physical_property, force_field_path, schema, options):
...
+                        continue
+
+                    new_list_value.extend(generated_path_list[value])
+
+                protocol.set_value(required_input, new_list_value)
+
+            # Update the schema of the modified protocol.
+            self.schema.protocols[protocol_id] = protocol.schema
 
     def _build_dependants_graph(self):
```
j-wags: Optional: "dependent" is the common spelling in U.S. English: https://grammarist.com/spelling/dependant-dependent/

propertyestimator/layers/simulation.py
```diff
@@ -540,7 +772,7 @@ def _execute_protocol(directory, protocol_schema, *parent_outputs, available_res
         try:
             output_dictionary = protocol.execute(directory, available_resources)
         except Exception as e:
-            # Except the unexpected...
+            # Except the unexcepted...
```
j-wags: Ha! Take care of yourself. You're no good to us dead :-P

propertyestimator/properties/enthalpy.py
```diff
@@ -2,154 +2,290 @@
...
+from propertyestimator.workflow.utils import ProtocolPath, PlaceholderInput
+
+
+@plugins.register_calculation_protocol()
+class WeightValueByMoleFraction(protocols.BaseProtocol):
+    """Multiplies a value by the mole fraction of a component
+    in a mixture substance.
+    """
+    @protocol_input(unit.Quantity)
+    def value(self, value):
```
j-wags: This is my first time digging into this convention. It's a little weird to me that we have decorators+functions being defined for what a (badly-abused) dict could do... I have to look especially closely at this in particular because it's is the interface that future developers will use when Simon transitions to a more science-based role. Maybe @andrrizzi could also give his thoughts? I won't make this a blocking review, but let's chat about this on our call Monday.

propertyestimator/substances.py
```diff
@@ -131,7 +131,7 @@ def __hash__(self):
 
         def __eq__(self, other):
 
-            return (self.smiles == other.self and
+            return (self.smiles == other.smiles and
```
j-wags: Let's start a conversation about where the SMILES come from. For this use case, we might toss in CMILES. It could help with some issues, like formatting alternating C-C single+double bonds into aromatic atoms and handling arbitrary protomers.

propertyestimator/tests/test_simulation_layer.py
```diff
@@ -0,0 +1,191 @@
...
+def test_calculation_schema(registered_property_name):
+    """Tests serialisation and deserialization of a calculation schema."""
+
+    registered_property = registered_properties[registered_property_name]
+
+    schema = registered_property.get_default_calculation_schema()
+    schema.validate_interfaces()
+
+    json_schema = schema.json()
+    print(json_schema)
```
j-wags: I generally try to avoid print statements in tests. They can make it tough to find errors/parse verbose pytest output.

propertyestimator/tests/test_workflow_protocols.py
```diff
@@ -0,0 +1,13 @@
+"""
+Units tests for propertyestimator.workflow
+"""
+
+def test_protocol_decorators():
```
j-wags: Is this a test will be implemented later, or should we scrap this file?

propertyestimator/workflow/groups.py
```diff
@@ -362,14 +390,17 @@ def _try_merge_protocol(self, other_protocol_id, other_group, merged_ids, parent
             raise RuntimeError('A protocol with id {} has already been merged '
                                'into the group.'.format(other_protocol_id))
 
-        protocols = self._root_protocols if parent_id is None else self._dependants_graph[parent_id]
+        protocol_ids = self._root_protocols if len(parent_ids) == 0 else []
```
j-wags: Will `parent_ids` be a single `str` (per docstring) or `List[str]`? This could be tough to debug if someone misinterprets, because both are index-able.

propertyestimator/workflow/groups.py
```diff
@@ -731,16 +836,31 @@ def merge(self, other):
...
             self.add_condition(condition)
 
         return merged_ids
 
-    def add_condition(self, condition):
+    def add_condition(self, condition_to_add):
 
-        if condition in self._conditions:
-            return
+        for condition in self._conditions:
```
j-wags: Add docstring

propertyestimator/workflow/protocols.py
```diff
@@ -377,6 +385,44 @@ def merge(self, other):
...
+        :obj:`ProtocolPath`.
+
+         Parameters
+         ----------
+         input_path: :obj:`ProtocolPath`
+            The input value to check.
+
+        Returns
+        -------
+        List[:obj:`ProtocolPath`]
```
j-wags: Oh cool. I bet this plays well with Sphinx -- Will need to check this out.

propertyestimator/layers/simulation.py
SimonBoothroyd: I think dependent is typically used as an adjective, and dependant is the noun - in this case because it's building a list of dependants (which are themselves dependent), I think dependant is correct? I'd be happy for this to go either way to be honest though, but this is probably beyond the scope of this PR as this is already established across the code base and should probably be proposed as a different PR?

propertyestimator/properties/enthalpy.py
SimonBoothroyd: 100% agree with the phrase badly-abused dict. A dict in this case would absolutely and unequivocally be a bad design choice here. These inputs and outputs are very much attributes of the class they exist in, and should be defined as such. Aside from this, putting these in a dict would be a use-ability and maintenance nightmare - off the top of my head quickly:

- for one there would be no IDE integration. When people are using these classes the IDE would not be able to suggest which inputs and outputs are available to get / set, making them a pain to use without having to dig through the docstring.
- you can't define doc strings for things in a dict. You would have to put them either in the __init__ docstring (bad choice) or in the class docstring where they might be hard to locate.
- all attributes would be have to be accessed by string literals both internally in the class and externally by the user. Not only does this result in really ugly looking code, it is just asking for typos leading to access bugs especially given that, as far as I'm aware, most static linters will not check if a key is present in a dict.
- It becomes awkward to tag attributes with meta, such as their anticipated types or their merge behavior, without either creating new dicts to store this in, or having the value dicts store class which have the value and meta data inside, which again will most likely lead to ugly code.

While the current descriptor design pattern may look a bit odd (there may be another way to clean this up), it avoids the above issues and gives a significant amount of control over how these attributes are `get` and `set` (such as validation type checking), allows easy metadata attachment, and in future will allow us to do some interesting, automatic stuff with docstrings.

propertyestimator/substances.py
SimonBoothroyd: +1 one for starting to think about this - great catch! Although when we switch over to the OFF toolkit V1.0.0 I think a lot of this functionality will be removed from the estimator, and simply pulled from your toolkit-independent api.

propertyestimator/tests/test_simulation_layer.py
SimonBoothroyd: In most cases I completely here, although here I've found it to be really helpful as when things fail, I can simply check the JSON to see what went wrong!

propertyestimator/tests/test_workflow_protocols.py
SimonBoothroyd: This will be a test implemented in my testing push next week (along with many others in this file), so it's probably safe to keep the file for now.

propertyestimator/workflow/groups.py
SimonBoothroyd: Typo - great catch!

propertyestimator/workflow/protocols.py
SimonBoothroyd: You'd think so... but at the moment doesn't!

propertyestimator/properties/enthalpy.py
andrrizzi: I'm a big fan of descriptors, and I never thought of using decorators to add the docstring!

I'd design this a little backwards though, to make this route of adding the descriptor with `__get__` and `__set__` possible
```python
class MyClass:
    value = ProtocolInput(unit.Quantity)
    value.__doc__ = "blabla"

    weighted_value = ProtocolOutput(unit.Quantity)
    weighted_value.__doc__ = "The value weighted by the `component`s mole fraction as determined from the `full_substance`."
```
which I think it's a slightly little more transparent. You can still have a decorator like
```python
def protocol_input(value_type, merge_behavior=MergeBehaviour.ExactlyEqual):
    def _protocol_input(cls_attribute):
        desc = ProtocolInput(value_tupe, merge_behavior)
        descr.__doc__ = cls_attribute.__doc__
        return descr
    return _protocol_input
```
Also, not sure if you're aware of the special function `__set_name__()`, which is pretty handy to determine the private name of the instance attribute in descriptors: https://stackoverflow.com/questions/42023852/how-can-i-get-the-attribute-name-when-working-with-descriptor-protocol-in-python

propertyestimator/properties/enthalpy.py
```diff
@@ -2,154 +2,290 @@
...
+        """
 
-        build_coordinates.substance = ProtocolPath('substance', 'global')
+        build_coordinates = protocols.BuildCoordinatesPackmol(id_prefix + 'build_coordinates')
 
-        schema.protocols[build_coordinates.id] = build_coordinates.schema
+        build_coordinates.substance = ProtocolPath('substance', 'global')
 
-        assign_topology = protocols.BuildSmirnoffTopology('build_topology')
+        assign_topology = protocols.BuildSmirnoffTopology(id_prefix + 'build_topology')
```
andrrizzi: What do you think about renaming this `BuildSmirnoffSystem`? The `Topology` object is a specific object in the basic `openforcefield` toolkit, and I don't think it has parameters so I'm worried this name could generate confusion.

propertyestimator/workflow/decorators.py
```diff
@@ -7,6 +7,8 @@
 
 from enum import Enum
 
+from propertyestimator.workflow.utils import PlaceholderInput
+
 
 class MergeBehaviour(Enum):
```
andrrizzi: I don't have a particular preference, but we probably need to settle on a single spelling convention. I think the US most common spelling is "behavior" in this case.

propertyestimator/properties/enthalpy.py
SimonBoothroyd: I agree!

propertyestimator/workflow/decorators.py
SimonBoothroyd: Definitely agree - good spot!

PR #13, opened 2019-02-09
Added the Napoleon Sphinx Extension

SimonBoothroyd:
## Description
Added the Napoleon sphinx extension to the conf.py file. This results in much cleaner looking docs pages, which properly pick up function parameter and property types, and link to their appropriate pages.

## Questions
- Does the type interpretation work consistently for both internal and external types?
- What is the correct way to specify the type in the Numpy docstring?

## Status
- [x] Ready to go


jchodera:
I'm not convinced we want to introduce a divergence in either our documentation or display standards for docstrings.

In the [openforcefield](https://github.com/openforcefield/openforcefield/blob/master/docs/conf.py#L40) repo, we use the [Sphinx `numpydoc` extension](https://github.com/numpy/numpydoc) which uses the [numpy docstring standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard). 

Is there a sufficiently compelling reason to need to change this for all our projects? If so, we should do this uniformly.

I realize our [current best practices docs](https://github.com/choderalab/software-development/blob/master/PYTHON_CODING.md#document-your-code) doesn't make this sufficiently clear, so we can certainly clarify that.

SimonBoothroyd:
> I'm not convinced we want to introduce a divergence in either our documentation or display standards for docstrings.

I 100% agree with this - at the moment I'm just using the PR to explore whether `sphinx.ext.napoleon` might be a better choice than `numpydoc`, and absolutely agree we shouldn't merge here unless we'll use it across the rest of the software stack!

> Is there a sufficiently compelling reason to need to change this for all our projects? If so, we should do this uniformly.

The main reason I was looking into `sphinx.ext.napoleon` is that `numpydoc` doesn't seem to have good support (I might have missed something and I'd be more than happy to eat my words here!) for cross-referencing function parameter types and property getter types. A [good example for this can be seen here](https://property-estimator.readthedocs.io/en/latest/api/generated/propertyestimator.layers.ReweightingLayer.html#propertyestimator.layers.ReweightingLayer.schedule_calculation), where ideally the majority of those parameter types should be links - I've tried a couple of different things but cross-referencing only seems to work for types which exist in the same module. The `sphinx.ext.napoleon` extension however seems to be able to pick all of these things up with no issues (for some reason RTD hasn't picked up this branch so I can't point to an example...), regardless of if we give full module paths or just class names.

I think cross-referencing of all types in an API doc is essential for ease of use, navigation and clarity, but other people might have different views on this?  Again it may be that I've missed something when trying to get this to work with `numpydoc`, and I'd be happy to be proved wrong here!

Other than that the two extensions give almost identical behavior, with both parsing the standard numpy docstrings. The only other difference is `sphinx.ext.napoleon` has a more compact (and in my opinion easier on the eyes) formatting [as can be seen here](https://stackoverflow.com/questions/36715801/difference-between-sphinxcontrib-napoleon-and-numpy-numpydoc).

j-wags:
I really like the look of Napoleon, and I think its implementation is orthogonal to numpydoc (that is, the two don't appear to conflict as far as I can tell).

The benefit is significant -- Getting live links between modules in the docs is something that I thought we already had. To show the difference in Simon's examples:

[Normal](https://property-estimator.readthedocs.io/en/latest/api/generated/propertyestimator.layers.ReweightingLayer.html#propertyestimator.layers.ReweightingLayer.schedule_calculation)

[Napoleon](https://property-estimator.readthedocs.io/en/napoleon_docs/api/generated/propertyestimator.layers.ReweightingLayer.html#propertyestimator.layers.ReweightingLayer.schedule_calculation)

@dgasmith, you're probably more experienced with RTD. Is there anything we need to watch out for if we use Napoleon?

andrrizzi:
If it supports the numpydoc standard and handles cross-referencing better it's probably worth trying it out. The changes required to make the substitution look very minimal judging from the PR.

jchodera:
I'm happy with either one you folks choose, but let's make sure we standardize on one choice and try to move everything (including the cookiecutter and our best practices docs) over!

jchodera:
> @jchodera, I think Napoleon will be a good choice for all of our repos, but I want Simon to bang on it for a few weeks as he writes his own docs, and be very vocal about the problems he encounters. If everything goes well, I'd be OK pushing Napoleon to all our repos and the cookiecutter in March.

Great! Please also update our best practices documentation [here](https://github.com/choderalab/software-development), which will serve as our de facto best practices guide until we can create an `openforcefield` guide.

PR #14, opened 2019-02-11
Finalize Property Estimator Public API

SimonBoothroyd:
## Description
A PR to finalize the design of the property estimators public API.

The property estimator public API is hopefully reaching (/ has hopefully reached) its final state (for the first cycle at least), and so it would be good to get as much feedback as possible on it now before anything is locked down. The key objects can be found in the `propertyestimator.client` and `propertyestimator.server` modules - these are namely `PropertyEstimatorClient`, `PropertyEstimatorServer`, `PropertyEstimatorOptions`, `PropertyEstimatorResult`.

Examples of how the property estimator will be mainly used can be either found in the `PropertyEstimatorClient`, `PropertyEstimatorServer` docs, or the examples folder.

## Questions

- What are you thoughts on the API design?
- Is there anything missing which you would wish to see when using the framework?

## Status
- [x] Ready to go


j-wags:
This pull request **introduces 7 alerts** when merging b4cc577fcce4f92b06247ab719bab49d7fc4f28f into d1b0c0b4c74d01e63e0660889cccd749e1a8c306 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-8af1da1d7e3a1ec64a373718a206136a8e09c49a)

**new alerts:**

* 6 for Unused import
* 1 for Unused local variable

---

docs/propertyestimator.rst
```diff
@@ -20,15 +20,15 @@ processes on a cluster:
...
+    estimator = PropertyEstimatorClient(nworkers=10) # NOTE: multiple backends will be supported in the future
     computed_properties = estimator.computeProperties(dataset, parameter_sets)
 
 Here, ``dataset`` is a ``PhysicalPropertyDataset`` or subclass, and ``parameter_sets`` is a list containing
 ``ForceField`` objects used to parameterize the physical systems in the dataset.
 
 This can be a single parameter set or multiple (usually closely related) parameter sets.
 
-``PropertyEstimator.computeProperties(...)`` returns a list of ``ComputedPhysicalProperty`` objects that provide access
+``PropertyEstimatorClient.computeProperties(...)`` returns a list of ``ComputedPhysicalProperty`` objects that provide access
```
andrrizzi: I can't find `computeProperties()` among the `PropertyEstimatorClient`'s methods. Is it outdated?

SimonBoothroyd: Thanks for the feedback @andrrizzi! I've incorporated most of this into the latest commit.

j-wags:
This pull request **introduces 6 alerts** when merging f62f6ccd6f98759641a7b5a2b181b51f5000e570 into d1b0c0b4c74d01e63e0660889cccd749e1a8c306 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-4d5374fa9eb4f758639e4f6ff95c724abcb95ebd)

**new alerts:**

* 6 for Unused import

---

docs/propertyestimator.rst
```diff
@@ -20,15 +20,15 @@ processes on a cluster:
 
 .. code-block:: python
 
-    estimator = PropertyEstimator(nworkers=10) # NOTE: multiple backends will be supported in the future
-    computed_properties = estimator.computeProperties(dataset, parameter_sets)
+    estimator = PropertyEstimatorClient(nworkers=10) # NOTE: multiple backends will be supported in the future
+    computed_properties = estimator.computeProperties(dataset, force_fields)
```
jchodera: Aren't we sticking to [PEP8](https://www.python.org/dev/peps/pep-0008/) convention here? That would be `estimator.compute_properties()` instead of `.computeProperties()`.

docs/propertyestimator.rst
```diff
@@ -45,7 +45,7 @@ This API can be extended in the future to provide access to the simulation data
 .. code-block:: python
 
     # Attach to my compute and storage resources
-    estimator = PropertyEstimator(...)
+    estimator = PropertyEstimatorClient(...)
```
jchodera: Does it make sense to call it `PropertyEstimatorClient()` if this is the end that the user sees? I definitely understand the idea of having a `PropertyEstimatorServer` class on the backend, but why not just call the client `PropertyEstimator` if this is what users will interact with?

We can also later move toward a usage model where, for simple calculations, the user doesn't even need to spin up a `PropertyEstimatorServer()` manually---the software can spin up a server on the current workstation if none is running or specified to connect to.

docs/propertyestimator.rst
SimonBoothroyd: This is perhaps best viewed on RTD: https://property-estimator.readthedocs.io/en/latest/propertyestimator.html

These pages are very out of date and should be showing a warning on the page! Most of this text was just integrated from the old The-PropertyEstimator-format.md, and is not reflective of the current API. When I get to my docs push later this week / early next week all of these will be fixed up!

docs/propertyestimator.rst
```diff
@@ -94,11 +94,11 @@ printed.
...
 
     # Load an initial parameter set
-    parameter_set = [ SMIRFFParameterSet('smarty-initial.xml') ]
+    force_field = [ SMIRFFParameterSet('smarty-initial.xml') ]
 
     # Compute physical properties for these measurements
-    estimator = PropertyEstimator(nworkers=10) # NOTE: multiple backends will be supported in the future
-    computed_properties = estimator.computeProperties(dataset, parameter_set)
+    estimator = PropertyEstimatorClient(nworkers=10) # NOTE: multiple backends will be supported in the future
+    computed_properties = estimator.computeProperties(dataset, force_field)
```
jchodera: Can `computeProperties()` really accept an iterable list of of `ForceField` objects?
If so, it would be great if it could also accept a single `ForceField` object so we don't need the awkward syntax of
```python
force_field = [ SMIRFFParameterSet('smarty-initial.xml') ]
computed_properties = estimator.computeProperties(dataset, force_field)
```
and can instead have
```python
force_field = SMIRFFParameterSet('smarty-initial.xml')
computed_properties = estimator.computeProperties(dataset, force_field)
```

examples/property_estimator/start_client.py
```diff
@@ -8,22 +8,43 @@
 from propertyestimator.utils import get_data_filename, setup_timestamp_logging
 
 
-def submit_calculation_to_server():
+def request_estimate_blocking():
```
jchodera: Suggest `compute_estimate()` might be simpler and clearer.
The blocking nature of this call is implied because we're asking for the estimate to be computed (now), as well as documented.

examples/property_estimator/start_client.py
```diff
@@ -8,22 +8,43 @@
...
     """Submit calculations to a running server instance"""
+    setup_timestamp_logging()
+
+    # Load in the data set of interest.
+    data_set = ThermoMLDataSet.from_file_list(get_data_filename('properties/single_density.xml'))
```
jchodera: Suggest `ThermoMLDataset.from_file_list` could just be called `.from_file()`, but accept either a single filename, an iterable of filenames, or a single `File`-like object (or iterable of `File`-like objects).

examples/property_estimator/start_client.py
```diff
@@ -8,22 +8,43 @@
...
+    # Load in the force field to use.
+    force_field = smirnoff.ForceField(get_data_filename('forcefield/smirnoff99Frosst.offxml'))
+
+    # Create the client object.
+    property_estimator = client.PropertyEstimatorClient()
+    # Submit the request to a running server, and wait for the results.
+    result = property_estimator.estimate(data_set, force_field)
```
jchodera: Is this the blocking- or non-blocking call? `estimate()` is good example of a clear, direct method name for a blocking call.

examples/property_estimator/start_client.py
```diff
@@ -8,22 +8,43 @@
...
 
+    logging.info('The server has returned a response: {}'.format(result))
+
+
+def request_estimate_non_blocking():
```
jchodera: Maybe the non-blocking call could be `compute_estimate_asynchronous()`, `compute_estimate_async()`, `estimate_asynchronous()`, or `estimate_async()`?

We might also draw some inspiration for idioms from the [Python asyncio module](https://docs.python.org/3/library/asyncio.html).

jchodera: Alternatively, you can just have an `.estimate()` method and add an optional argument `asynchronous=False` by default, with the user able to specify `asynchronous=True` to get an asynchronous estimate.

examples/property_estimator/start_client.py
```diff
@@ -8,22 +8,43 @@
...
     data_set = ThermoMLDataSet.from_file_list(get_data_filename('properties/single_density.xml'))
+    # Load in the force field to use.
     force_field = smirnoff.ForceField(get_data_filename('forcefield/smirnoff99Frosst.offxml'))
 
-    property_estimator = client.PropertyEstimator()
-    ticket_ids = property_estimator.submit_computations(data_set, force_field)
+    # Create the client object.
+    property_estimator = client.PropertyEstimatorClient()
+    # Submit the request to a running server.
+    ticket_id = property_estimator.estimate(data_set, force_field)
```
jchodera: I thought we had the idea to move away from returning `ticket_id`s, instead returning a `PropertyEstimate` or `PropertyEstimatResult` result object that either contained the resulting estimate (after a synchronous call) or knew how to query/retrieve the asynchronous estimate (after an async call). Or maybe it could even retrieve the estimate in the background, or when first queried for the result.

propertyestimator/datasets/datasets.py
```diff
@@ -139,3 +139,26 @@ def filter_function(x):
...
+        number_of_components : int
+            The allowed number of components in the mixture.
+
+        Examples
+        --------
+        Filter the dataset to only include pure substance properties.
+
+        >>> # Load in the data set of properties which will be used for comparisons
+        >>> from propertyestimator.datasets import ThermoMLDataSet
+        >>> data_set = ThermoMLDataSet.from_doi_list('10.1016/j.jct.2016.10.001')
```
jchodera: Suggest `ThermoMLDataset.from_doi()` would be simpler, and could either take a DOI or an iterable of DOIs.

docs/propertyestimator.rst
SimonBoothroyd: I personally prefer it with Client because it makes it more obvious that this isn't just built around one class on one machine, but rather this will simply be the class used to interact with the main compute system. I believe this is the naming convention @dgasmith went with for QCFractal. Id be happy to drop the name though if people don't agree!

docs/propertyestimator.rst
SimonBoothroyd: Unfortunately this is out of date docs again like in the previous comment - they will be updated in the near future.

examples/property_estimator/start_client.py
SimonBoothroyd: Yeah I think what you suggest looks better.

docs/propertyestimator.rst
dgasmith: `DaskClient`, `MongoClient`, etc common and explicit.

docs/propertyestimator.rst
jchodera: Also, I think the API for this has changed. I see this was from my old notes/sketch!

examples/property_estimator/start_client.py
SimonBoothroyd: The synchronous= route is the way I updated it to a few minutes ago per @andrrizzi's suggestion, and I think that looks pretty clean.

propertyestimator/client.py
```diff
@@ -125,39 +143,34 @@ class Config:
         }
 
 
-class PropertyEstimator(object):
-    """The main object responsible for requesting a set of properties
-    be calculated by the low-level property calculation backend.
+class PropertyEstimatorClient:
```
jchodera: Suggest `PropertyEstimator` instead of `PropertyEstimatorClient`, since this is the user-facing end.

propertyestimator/client.py
```diff
@@ -125,39 +143,34 @@ class Config:
...
-    >>>
-    >>> # Instruct the server to listen for incoming submissions
-    >>> property_server.run_until_killed()
+    >>> from propertyestimator.client import PropertyEstimatorClient
+    >>> property_estimator = PropertyEstimatorClient()
 
-    If the server and client initialisation occur in the same function call (e.g. the same
-    main function), then `property_server.run_until_complete()` must be called after
-    all computations have been submitted.
+    If the :obj:`PropertyEstimatorServer` is not running on the local machine, you will
```
jchodera: If no `PropertyEstimatorServer` is found running locally or specified in the constructor, would it be possible for the `PropertyEstimator` (`PropertyEstimatorClient`) to create a `PropertyEstimatorServer` with simple `DaskLocalClusterBackend(ncpus)` and `LocalFileStorage()` compute and storage backends by default if none is specified?

propertyestimator/client.py
```diff
@@ -170,40 +183,48 @@ class PropertyEstimator(object):
     >>>
     >>> # Load initial parameters
     >>> from openforcefield.typing.engines.smirnoff import ForceField
-    >>> parameters = ForceField('smirnoff99Frosst.offxml')
+    >>> parameters = ForceField(['smirnoff99Frosst.offxml'])
     >>>
-    >>> # Create a property estimator
-    >>> from propertyestimator.client import PropertyEstimator
-    >>> property_estimator = PropertyEstimator()
+    >>> ticket_id = property_estimator.estimate(data_set, parameters)
```
jchodera: I think we had wanted to always return a `PropertyEstimatorResult` instead of `ticket_id`s, since the `PropertyEstimatorResult` would be able to have code to manage either asyncrhonous or synchronous results.

SimonBoothroyd: So this is something I've been conflicted about when I started looking at implementing this - In principle it seems like a really elegant solution, but in practice I'm not sure. What if for example you submit some jobs to the server, then your laptop dies. Or you submit some jobs and the server you're submitting to is super busy and can't return your results for a few day / a week etc... That PropertyEstimatorResult object or Future or whatever the return was is now lost, and you would have difficulty knowing how to ask the server for those results now.. With a simple string identifier, one can submit the jobs, wait however long it takes to get the results back, and then easily query the server about them without having to have kept the `Future` or result object alive.

SimonBoothroyd: @dgasmith do you have any thoughts about this? I'm sure you must have already had to make this design decision?

dgasmith: Tough call, in general even returning an ID isn't enough unless you snapshot that ID to disk somewhere. Can your clients recover state if they die?

SimonBoothroyd: They can't at the moment.

SimonBoothroyd: Sorry this is probably a stupid question, but what do you mean by recover state?

dgasmith: ForceBalance is running along and it requests a new job and then promptly dies. When ForceBalance recovers does it know anything about the jobs that it submitted before it died? If yes, then keeping the id's around is a good id. If no, it is a bit of a moot point.

SimonBoothroyd: I'm not 100% sure what ForceBalance is doing under the hood at the moment, but I'd argue that it should be saving the state of what its waiting for (so the id's). If I was to submit a large benchmark for example, I would definitely be wanting to save the state so that I could come back in a week and recover it.

propertyestimator/client.py
SimonBoothroyd: I've been struggling to think of a clean way to do this, but I think it should actually be possible - I think if we set up a PropertyEstimatorServer cli as an entrypoint, we could just `nohup` it from within python if no server address is given and no instance is running. We would then probably track the pid of the nohub, and make sure we kill it when we exit the method. The only issue with this is that if something goes wrong and the program doesn't exit cleanly, the server would be left running perhaps without the users knowledge..

SimonBoothroyd: The only other issue that may pop up (and I don't know if this would ever actually happen) might be that you call this method, and it spins up a local server. You may then call this method again from another script on the same machine. This new script would see the spun up server and launch its jobs on there. The original script may then finish and kill the server, before the jobs submitted by the second script have finished.

propertyestimator/client.py
jchodera: Could this be resolved by simply making `PropertyEstimatorResult` serializable?

propertyestimator/client.py
jchodera: That could be resolved by simply having two behaviors:
* If no server is specified, spin up a local server that the client owns (and kills when the client dies)
* If a server is specified, use that (but don't kill it when the client dies)

propertyestimator/client.py
leeping: At the moment, ForceBalance does not know anything about the jobs that it submitted before it died.  The remote jobs that run individual simulations at different thermodynamic phase points are "workers" that connect to FB that acts as the "server".  

FB has a checkpointing system at the level of individual targets and optimization cycles.  That is to say, it can restart an optimization from saved values of the parameters, and if the latest optimization cycle has not finished evaluating the objective function, it will read the calculation results from targets where the calculations have been completed.  

If the server allows different processes to query running jobs and retrieve results, I believe it should have nice benefits but I'm not sure exactly what they would be.

propertyestimator/client.py
andrrizzi: Another solution could be adding a flag to the `PropertyEstimatorServer` that makes it self-kill after a period of inactivity and have the `PropertyEstimatorClient` always turn on that flag when automatically creating a server.

propertyestimator/client.py
andrrizzi: A clarification, what is the role of `PropertyEstimatorResult.id`? Is it the same as `ticket_id` in this case?

SimonBoothroyd: So in both cases the id referred to is the id which the server assigns to the request. The two should be the same.

@andrrizzi, @jchodera, @j-wags I've just committed a slight refactor which tries to implement the design pattern of returning an object which itself can be asked to either synchronously or asynchronously return the results (kind of similar to returning a Future). Would you be able to take a look and let me know if this is closer to what you had in mind? The `estimate` method has been renamed to `request_estimate` and no longer takes a `synchronous` argument, as this is now delegated to the returned `Request` object. The `request_estimate` simply sends the request to the server object.

SimonBoothroyd: I'm not 100% sure if I like it more than just returning an id, but I'd be happy to go with this pattern if the consensus is that it's better!

SimonBoothroyd: I think this design pattern makes spinning up a server instance if one isn't running more challenging. Because the `request_estimate` method no longer has control over when the data is being returned, it won't know when to kill the server. While we could do as @andrrizzi  suggest and kill it if it's been inactive for a while, this would lead to issues if the user has waited too long to request the results of the estimate!

andrrizzi:
I like it! I've seen this pattern of returning a request object that can be queried later in other libraries as well so it should be pretty clear to the user what is going on underneath by just looking at the examples.

> While we could do as @andrrizzi suggest and kill it if it's been inactive for a while, this would lead to issues if the user has waited too long to request the results of the estimate!

Good point! In general, I think we need to define what is the mechanism we want to implement to reduce the impact of failures. Maybe we can postpone the decision of whether to spin up automatically a local server for later in a separate thread since, on first thinking, I don't think it will require API breaks, and it's a use case that will be useful mostly only for testing/debugging.

SimonBoothroyd:
Ok I think this PR now contains everyones suggestions - if no one has any further changes they wish to see can someone LGTM and approve this?

PR #15, opened 2019-02-15
First implementation of density reweighting

SimonBoothroyd:
## Description
I have finished a first pass at roughly implementation for allowing reweighting of densities. This required quite large changes to parts of the storage system - in particular to add support for serializing data for properties such as enthalpy of mixing.

For now most of the reweighting implementation lives in the `density.py` class. While this might be ok (but not the best way) for a small number of built in properties (i.e reweighting implementation in the class), this needs to be assessed and redesigned to be more modular and flexible in the future.

This PR also mostly fixes up the enthalpy_of_mixing implementation, adding support for storing all of the pure component trajectory data.

## Status
- [ ] Ready to go


j-wags:
This pull request **introduces 3 alerts** when merging 73febe2cf8dc2cfd6d909a8e7710ed9d07759b35 into d1b0c0b4c74d01e63e0660889cccd749e1a8c306 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-ace995316f62d2141551929056dcadeb4780c49d)

**new alerts:**

* 2 for Unused import
* 1 for Incomplete ordering

---

propertyestimator/layers/reweighting.py
```diff
@@ -42,24 +90,188 @@ def schedule_calculation(calculation_backend, storage_backend, layer_directory,
...
+        target_force_field_id: str
+            The id of the target force field parameters.
+        target_force_field: ForceField
+            The target set of force field parameters.
+        existing_data: StoredSimulationData
+            The reference data to get the energies from.
+
+        Returns
+        -------
+        :obj:`numpy.ndarray`
```
jchodera: What dimensionality is the array? Is it something like `(nsnapshots)` in dimension?

propertyestimator/layers/reweighting.py
```diff
@@ -42,24 +90,188 @@ def schedule_calculation(calculation_backend, storage_backend, layer_directory,
...
+
+        if target_force_field_id != existing_data.parameter_set_id:
+
+            potential_energies = ReweightingLayer.resample_data(substance, existing_data.trajectory_data,
+                                                                target_force_field, available_resources)
+
+        reduced_potentials = []
+
+        for index, potential_energy in enumerate(potential_energies):
+            reduced_potentials.append(potential_energy / unit.kilojoule_per_mole)
```
jchodera: We should never silently use a specific arbitrary unit like this. There are only two ways to store dimensionless quantities, and we need to be clear about documenting them:
* The [reduced potential](http://www.alchemistry.org/wiki/Multistate_Bennett_Acceptance_Ratio#Reduced_potential) is the dimensionless ratio `potential_energy / kT`, not the value in kJ/mol. It is recommended we use this definition for potential energies whenever reweighting or statistical mechanics is performed with the energies, since the probability density is just `p(x) \propto \exp[-u(x)]`, which makes this simple to work with.
* For all other cases---including when estimating average enthalpies or energies using observables---if we need to store dimensionless quantities, we should use `quantity.value_in_unit_system(unit_system)` where we standardize the unit system to be, for example, [SI units](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/unit/unit_definitions.py#L308-L316).

propertyestimator/layers/reweighting.py
```diff
@@ -42,24 +90,188 @@ def schedule_calculation(calculation_backend, storage_backend, layer_directory,
...
+
+        # If the parameters haven't changed, we only need to swap out the T / P?
+        potential_energies = existing_data.statistics_data.get_observable(ObservableType.PotentialEnergy)
+
+        if target_force_field_id != existing_data.parameter_set_id:
+
+            potential_energies = ReweightingLayer.resample_data(substance, existing_data.trajectory_data,
+                                                                target_force_field, available_resources)
+
+        reduced_potentials = []
```
jchodera: It's generally more convenient to manipulate numpy arrays.

If `potential_energies` is a numpy array, this code cam be compressed to simply
```python
reduced_potentials = potential_energies / kT
```

propertyestimator/layers/reweighting.py
```diff
@@ -42,24 +90,188 @@ def schedule_calculation(calculation_backend, storage_backend, layer_directory,
...
+        :obj:`numpy.ndarray`
+            The reduced potential of the target state.
+        """
+
+        # If the parameters haven't changed, we only need to swap out the T / P?
+        potential_energies = existing_data.statistics_data.get_observable(ObservableType.PotentialEnergy)
+
+        if target_force_field_id != existing_data.parameter_set_id:
+
+            potential_energies = ReweightingLayer.resample_data(substance, existing_data.trajectory_data,
```
jchodera: Most of the code to do this robustly and expressively already exists. Suggests you might look carefully at @andrrizzi's [`openmmtools.states.ThermodynamicStates`](https://openmmtools.readthedocs.io/en/0.17.0/api/generated/openmmtools.states.ThermodynamicState.html#openmmtools.states.ThermodynamicState) which does most everything you need for computing reduced potentials correctly with OpenMM.

I'd also highly suggest you integrate the [`openmmtools.mcmc`](https://openmmtools.readthedocs.io/en/0.17.0/mcmc.html) module to do all your sampling, and use [`openmmtools.states.SamplerState`](https://openmmtools.readthedocs.io/en/0.17.0/api/generated/openmmtools.states.SamplerState.html#openmmtools.states.SamplerState) to manipulate and store any snapshot data. This would allow us to do more clever things with sampling in the future without needing to make major changes to the code, since we can simply compose MCMC blocks for this purpose.

If you find anything missing from [`openmmtools`](https://openmmtools.readthedocs.io/en/0.17.0/), feel free to raise an issue, or even contribute! We should try to move any generally reusable code---maybe even your pipeline code---there if there's a chance it could be of use to multiple projects!

propertyestimator/layers/reweighting.py
```diff
@@ -42,24 +90,188 @@ def schedule_calculation(calculation_backend, storage_backend, layer_directory,
...
+            return PropertyEstimatorException(directory='',
+                                              message='Failed to create a system from the'
+                                                      'provided topology and molecules')
+
+        context = None
+        integrator = VerletIntegrator(0.002 * unit.picoseconds)
+
+        if available_resources.number_of_gpus > 0:
+
+            gpu_platform = Platform.getPlatformByName('CUDA')
```
jchodera: We should never hardcode the `CUDA` platform, since it will lead to exceptions if the wrong CUDA toolkit version is installed or no GPUs are installed. If we don't specify a platform, we can instead fall back gracefully to `OpenCL` or `CPU` platforms. It's a bit more complicated to create a platform and make sure we're using both deterministic PME and mixed precision context properties, since these aren't the same on all platforms, but @andrrizzi has essentially hidden all of this detail for us under the hood of his [`ContextCache`](https://openmmtools.readthedocs.io/en/0.17.0/api/generated/openmmtools.cache.ContextCache.html#openmmtools.cache.ContextCache) and [`ThermodynamicState`](https://openmmtools.readthedocs.io/en/0.17.0/api/generated/openmmtools.states.ThermodynamicState.html#openmmtools.states.ThermodynamicState) objects, which we should use whenever possible.

jchodera:
In particular, I bet there's a way to greatly simplify some of the `Protocol` code by using [`openmmtools.mcmc`](https://openmmtools.readthedocs.io/en/0.17.0/mcmc.html).

I think there's also an opportunity to simplify the handling of `Ensemble`s by extending the API of [`ThermodynamicState`](https://openmmtools.readthedocs.io/en/0.17.0/api/generated/openmmtools.states.ThermodynamicState.html#openmmtools.states.ThermodynamicState) to know what thermodynamic ensemble it defines. In essence, it already knows this, since it knows how to compute the reduced potential from relevant thermodynamic parameters and dynamical variables---which is all that really matters---but it could expose some useful additional API points to return human-interpretable ensembles (e.g. NVT, NPT, µVT) if that would be helpful.

propertyestimator/workflow/groups.py
```diff
@@ -749,6 +753,12 @@ def _evaluate_condition(self, condition_type, left_hand_value, right_hand_value)
         if left_hand_value is None or right_hand_value is None:
             return False
 
+        if isinstance(left_hand_value, EstimatedQuantity):
+            left_hand_value = left_hand_value.uncertainty
+            
+        if isinstance(right_hand_value, EstimatedQuantity):
+            right_hand_value = right_hand_value.uncertainty
```
andrrizzi: Should the condition be explicit about whether the user want to specify an uncertainty or an estimated quantity instead? For example, what if I want to check something like `EstimatedQuantity(1.2*kcal/mol, uncertainty=2.0*kcal/mol) < 1.5*kcal/mol`, which, if we're interested in the statistical significance, would return `False`?

propertyestimator/layers/reweighting.py
```diff
@@ -42,24 +90,188 @@ def schedule_calculation(calculation_backend, storage_backend, layer_directory,
...
+
+                with open(data_path, 'rb') as file:
+                    existing_data[substance_id].append(pickle.load(file))
+
+        for force_field_id in existing_force_field_paths:
+
+            with open(existing_force_field_paths[force_field_id], 'rb') as file:
+                existing_force_fields[force_field_id] = deserialize_force_field(pickle.load(file))
+
+        reweighted_property = property_class.reweight(physical_property, options, parameter_set_id, existing_data,
```
andrrizzi: If `reweight` needs to access instance attribute, should it be an instance method rather than a `staticmethod`?

jchodera:
Also, it would be great to eventually migrate the `Protocol` code into `openmmtools` if it looks like a flexible way to build simulation pipelines and do distributed computing with them!

j-wags:
This pull request **introduces 3 alerts** when merging e1dfbb73cb4b29a6a3fc3012974d5e92477aa7e5 into d1b0c0b4c74d01e63e0660889cccd749e1a8c306 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-9a4df899711e24897983eccb847ff7bea8c555bb)

**new alerts:**

* 2 for Unused import
* 1 for Incomplete ordering

---

propertyestimator/layers/reweighting.py
SimonBoothroyd: It is - I've added this to the docstring.

propertyestimator/layers/reweighting.py
SimonBoothroyd: Good catch - I'm not sure why this was added as it redundant in this case anyway!

propertyestimator/utils/statistics.py
```diff
@@ -179,28 +228,49 @@ def from_pandas_csv(cls, file_path):
...
-        data = np.array([
-            self._internal_data[AvailableQuantities.PotentialEnergy] / unit.kilojoule_per_mole,
-            self._internal_data[AvailableQuantities.KineticEnergy] / unit.kilojoule_per_mole,
-            self._internal_data[AvailableQuantities.TotalEnergy] / unit.kilojoule_per_mole,
-            self._internal_data[AvailableQuantities.Temperature] / unit.kelvin,
-            self._internal_data[AvailableQuantities.Volume] / unit.nanometer**3,
-            self._internal_data[AvailableQuantities.Density] / unit.gram * unit.milliliter,
-            self._internal_data[AvailableQuantities.Enthalpy] / unit.kilojoule_per_mole])
+    @classmethod
+    def from_existing(cls, existing_instance, data_indices=None):
```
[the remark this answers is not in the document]
SimonBoothroyd: Good call!

propertyestimator/workflow/groups.py
SimonBoothroyd: So theres two parts to this and neither are particularly satisfactory unfortunately..

At the moment we have to hard code in that if we have passed in an estimated quantity we only look at the uncertainty because the ProtocolPath class is not generic enough at the moment to be able to point to properties within classes - but this is coming in the future.

The other thing is for now I've disabled <, > checks on EstimatedQuantity until I can think them through, because I'm not 100% sure what the expected behaviour of them is (and same with ==) should we be comparing whether one thing is significantly (in a statistical sense) less than another thing, or should we check whether one thing is less than the other thing to within machine precision.

propertyestimator/layers/reweighting.py
SimonBoothroyd: I agree - I'm not sure why I added these things in as lists initially, but I've refactored so things are passed around as ndarrays now.

propertyestimator/layers/reweighting.py
SimonBoothroyd: In this case I'd argue we definitely do want the exception to be raised. When you set up the `PropertyEstimatorServer` you specify which compute resources (number of cores, gpus etc) you want each worker to have, with the expectation that the worker will make full use of these resources when possible. We don't want openmm (and thus the estimator) to silently fall back to using threads because of a bad cuda installation, or because there are no GPUs as this would be contrary to the expected user experience, and we should inform the user that what they are asking for could not be completed so they can remedy it.

SimonBoothroyd: I'll definitely add a flag to the resources object however to let the user choose between CUDA and OpenCL.

jchodera: You're right about not wanting to fall back to CPU threads, but the CUDA version match requirements are so finicky that I *do* think we want to be able to fall back to OpenCL if CUDA is unavailable. 

We also *definitely* want to make sure the right platform parameters get set to ensure we have deterministic forces and mixed precision.

And in general, instead of hardcoding a lot of this general functionality, it would be great to encapsulate it in general management helper classes like [`ContextCache`](https://openmmtools.readthedocs.io/en/0.17.0/api/generated/openmmtools.cache.ContextCache.html#openmmtools.cache.ContextCache).

SimonBoothroyd: I think what you propose is a good way forward! I'm working on refactoring the reweighting code to be more modular and to take advantage of the protocol workflows and the openmmtools utilities, so will make sure to include these changes also.

propertyestimator/workflow/groups.py
andrrizzi: > I've disabled <, > checks on EstimatedQuantity until I can think them through, because I'm not 100% sure what the expected behaviour of them is

I see. That probably deserves its own discussion/PR.

j-wags:
This pull request **introduces 2 alerts** when merging 8559498eb7647d877eef8da308bddda695f7112a into 9424366d119cad0c35127dedebfb1bf70fe77bb8 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-9b2f7641676713f66e75368445242b4b6d0ca4e6)

**new alerts:**

* 1 for Unused local variable
* 1 for Unused import

---

j-wags:
This pull request **introduces 1 alert** when merging 6cb57c6d918b92f446153926b497ff9ad4385bc3 into 9424366d119cad0c35127dedebfb1bf70fe77bb8 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-2390e156b1b620c7f6b44c3daa85c635c64336cc)

**new alerts:**

* 1 for Unused import

---

SimonBoothroyd:
I've gone ahead and implemented the majority of the suggestions. Once the travis and lgtm checks have completed, I'm going to merge and close this PR. If any one has any other feedback feel free to drop it here or in an issue and I'll open a new PR to address to save this PR getting even more bloated.

j-wags:
This pull request **introduces 4 alerts** when merging 02c8dc927c8984099c8f6261414729dbe8117da5 into 9424366d119cad0c35127dedebfb1bf70fe77bb8 - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-b898ef7748f6b05dfbccfa45fd040e9e7c90c073)

**new alerts:**

* 4 for Unused import

---

PR #23, opened 2019-04-15
Adding Dask LSF Job Queue Backend for better HPC Integration

SimonBoothroyd:
## Description
This PR adds support for the dask-jobqueue LSFCluster compute backend, which will enable better support for the propertyestimator on the MSKCC lilac cluster (and any other cluster which uses the LSF queueing system), and will serve as a test case before expanding to support to other queue systems such as SLURM and PBS (also the dask-jobqueue)

## Todos
  - [x] Create a DaskLSFBackend object
  - [x] Add a way for the workflow graph to check if a protocol has already been run
  - [x] Add a checkpointing system for `ConditionalGroup` protocols
  - [x] Switch the RunOpenMMSimulation protocol to use checkpoint files (optional)

## Status
- [x] Ready to review
- [ ] Ready to go


j-wags:
This pull request **introduces 1 alert** when merging c8bdc00d0eb21cf11d2b98ee5ec852447ee08de4 into 9319f0d645eed6603cbe567c620ad0fe5da590de - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/propertyestimator/rev/pr-831820c8a3a9c5871758b82ef87c6faf9612e145)

**new alerts:**

* 1 for Unused import

---

propertyestimator/backends/dask.py
```diff
@@ -1,13 +1,228 @@
...
+        resources_per_worker: QueueWorkerResources
+            The resources to request per worker.
+        default_memory_unit: simtk.Unit
+            The default unit used by the LSF queuing system when
+            defining memory usage limits / requirements - this
+            must be compatible with `unit.bytes`.
+        queue_name: str
+            The name of the queue which the workers will be requested
+            from.
+        extra_script_commands: list of str
```
andrrizzi: I'd suggest the name `setup_script_commands`. It's probably more intuitive when you're familiar with tests and/or `timeit`.

propertyestimator/backends/dask.py
```diff
@@ -1,13 +1,228 @@
...
+        # to multiply by a corrective factor to remove this assumption.
+        lsf_byte_scale = (1 * (unit.mega * unit.byte)).value_in_unit(self._default_memory_unit)
+        memory_bytes *= lsf_byte_scale
+
+        job_extra = None
+
+        if self._resources_per_worker.number_of_gpus > 0:
+
+            job_extra = [
+                '-gpu num={}:j_exclusive=yes:mode=shared:mps=no:'.format(self._resources_per_worker.number_of_gpus)
```
andrrizzi: I'm not sure how `self._resources_per_worker.number_of_gpus` should be interpreted, but I want to make sure that you are aware that with IBM LSF syntax (the one installed on Lilac), `-gpu num=4` means 4 GPUs _per node_.

propertyestimator/workflow/protocols.py
```diff
@@ -992,14 +1012,25 @@ def _setup_new_simulation(self, directory, temperature, pressure, available_reso
                                     integrator,
                                     platform)
 
-        box_vectors = input_pdb_file.topology.getPeriodicBoxVectors()
+        checkpoint_path = path.join(directory, 'trajectory.chk')
```
andrrizzi: I don't think the `CheckpointReporter` stores the full trajectory. I believe only the last snapshot is saved on disk (and overwrites the old one).

SimonBoothroyd: Good catch - this file was meant to just be called something like `checkpoint.chk` but didn't get changed when copy pasting from above!

propertyestimator/backends/dask.py
SimonBoothroyd: > I'm not sure how `self._resources_per_worker.number_of_gpus` should be interpreted

This whole object describes the resources which will be requested per dask worker (i.e per each submission which is made to the LSF queue) such that `number_of_gpus` is the number of gpu's which will be requested per worker. It is expected that the resources requested here should fit within a single node (unfortunately there is no simple way to validate these that I can see before a worker is requested) - I'll add a note to clarify this.

> I want to make sure that you are aware that with IBM LSF syntax (the one installed on Lilac), `-gpu num=4` means 4 GPUs _per node_.

Thanks for pointing this out - this way my understanding also!

PR #27, opened 2019-04-25
Dask Improved Support / Bug Fixes

SimonBoothroyd:
## Description
A PR to fix some subtle bugs which have been observed while testing the dask LSF backend and general QoL changes to aid in debugging.

## Todos
  - [x] Increase the number of times tasks can be relaunched on queue backends
  - [x] Give meaningful names to dask tasks to aid in debugging

## Status
- [x] Ready to go

### DIFF
--- a/propertyestimator/backends/dask.py
+++ b/propertyestimator/backends/dask.py
@@ -205,7 +205,7 @@ class DaskLSFBackend(BaseDaskBackend):
         # exceptions up as PropertyEstimatorExceptions and return these
         # gracefully (such that the task won't be marked as failed by
         # dask).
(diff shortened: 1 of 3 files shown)


propertyestimator/backends/dask.py
```diff
@@ -177,8 +198,14 @@ def __init__(self,
             if resources_per_worker.number_of_gpus > 1:
                 raise ValueError('Only one GPU per worker is currently supported.')
 
-        self._cluster = None
-        self._client = None
+        # For now we need to set this to some high number to ensure
```
j-wags: This is a good detail to comment on, and I'm sure this will help new people coming into the codebase.

PR #30, opened 2019-05-06
Best practices integration

ocmadin:
## Description
Move new best practices documentation format into main branch

## Questions
- [ ]  Going forward, what's the best way to make changes to the documentation?  I have been editing the .rst files directly on github, but if there's a better way of doing it (that also doesn't create so many commits) I'm open to it

## Status
- [ ] Ready to go!


SimonBoothroyd:
Looks good to me! I don't know if we want to develop this a bit further in this branch before merging fully into master?

> what's the best way to make changes to the documentation

I normally edit files in PyCharm and then just commit / push them to github in larger batches to avoid lots of small changes. 

If you want we can chat about building the docs locally as well so you can instantly see changes rather than waiting for read the docs to build!

ocmadin:
> Looks good to me! I don't know if we want to develop this a bit further in this branch before merging fully into master?

That was my thought, too - but the consensus on Thursday's call was for me to create a PR, so I did.  Can you preview markdown files in PyCharmm?

SimonBoothroyd:
Sounds good!

It has semi-decent support for previewing the files although it doesn't always 100% match what will show on RTD, but it offers a good idea of what things will look like.

I added a basic guide on how to build the docs on your local machine [here](https://property-estimator.readthedocs.io/en/best_practices_integration/install.html#building-the-documentation-locally), which will look identical to those that will be generated on RTD for easier preview without waiting for RTD to build. 

Once you have have the env setup, you just need to run

```rm -rf api && make clean && make html```

In the docs directory and it will update the docs and a few seconds!

jchodera:
@SimonBoothroyd docs builds are failing:
image
Looks like the environment installation is either failing or timing out.

PR #34, opened 2019-05-09
Initial Yank Host-Guest Binding Affinity Integration

SimonBoothroyd:
## Description
A PR to implement direct host-guest binding affinity calculations into the estimator framework. The bulk of this PR has been to introduce two new protocols:

 - A `BaseYankProtocol` which handles the actual running of yank calculations, setting up base input `yaml` files and will allow easy integration of hydration / solvation free energy calculations
- A `LigandReceptorYankProtocol` which extends the base input `yaml` file creation to include specific options for ligand-receptor binding calculations.

This PR also fixes a number of the `packmol` utility bugs, as well as extends it to support solvation of structures.

This PR also comes with the usual plethora of bug fixes / QoL improvements

## Todos
  - [x] Test initial set up pipeline.

## Status
- [x] Ready to go

### DIFF
--- a/propertyestimator/protocols/forcefield.py
+++ b/propertyestimator/protocols/forcefield.py
@@ -5,6 +5,8 @@ A collection of protocols for assigning force field parameters to molecular syst
 import logging
 from os import path
 
+import numpy as np
+from simtk import unit
 from simtk.openmm import app
 
 from propertyestimator.substances import Substance
@@ -30,6 +32,15 @@ class BuildSmirnoffSystem(BaseProtocol):
         force field parameters will be assigned."""
         pass
 
+    @protocol_input(dict)
+    def charged_molecule_paths(self):
+        """File paths to mol2 files which contain the charges assigned to molecules
+        in the system. This input is helpful when dealing with large molecules (such
+        as hosts in host-guest binding calculations) whose charges may by needed
+        in multiple places, and hence should only be calculated once. The key should
+        be the smiles or unique label of the molecule"""
+        pass
+
     @protocol_input(Substance)
     def substance(self):
         """The composition of the system."""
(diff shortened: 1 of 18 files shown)


propertyestimator/protocols/simulation.py
```diff
@@ -343,7 +385,124 @@ def _get_yank_options(self):
...
+
+            'minimize': True,
+
+            'default_number_of_iterations': self._number_of_iterations,
+            'default_nsteps_per_iteration': self._steps_per_iteration,
+            'checkpoint_interval': self._checkpoint_interval,
+
+            'default_timestep': quantity_to_string(self._timestep),
+
+            # TODO: Are these always sensible choices? I guess
```
jchodera: Yes, these are generally sensible choices!

propertyestimator/protocols/simulation.py
```diff
@@ -353,34 +512,117 @@ def execute(self, directory, available_resources):
...
+
+        self._ligand_residue_name = None
+
+        self._solvated_ligand_coordinates = None
+        self._solvated_ligand_system = None
+
+        self._solvated_complex_coordinates = None
+        self._solvated_complex_system = None
+
+        self._local_ligand_coordinates = 'ligand.pdb'
```
jchodera: If you're using the Yank Python API, you shouldn't need to go through `.pdb` and `.xml` files---unless this is more convenient!
http://getyank.org/latest/api/yank_api.html

propertyestimator/protocols/forcefield.py
```diff
@@ -49,9 +83,47 @@ def __init__(self, protocol_id):
...
         self._system_path = None
 
+    @staticmethod
+    def _generate_known_charged_molecules():
+        """Generates a set of molecules whose charges are known a priori,
+        such as ions, for use in parameterised systems.
+
+        Notes
+        -----
+        These are solely to be used as a work around until library charges
```
jchodera: Maybe leave a TODO to remove this when library charges are added?

propertyestimator/protocols/simulation.py
SimonBoothroyd: For now this is more convenient for how things are set up as a whole - but I'll look into switching over to this in future!

propertyestimator/protocols/simulation.py
```diff
@@ -332,57 +346,477 @@ class BaseYankProtocol(BaseProtocol):
...
+        ----------
+        checkpoint_path: str
+            The path to the yank `.nc` file
+        output_trajectory_path: str
+            The path to store the extracted trajectory at.
+        """
+
+        from yank.analyze import extract_trajectory
+
+        mdtraj_trajectory = extract_trajectory(checkpoint_path, state_index=0)
```
andrrizzi: A couple of suggestions:
- If this is a periodic simulation (I think it is), I'd set `image_molecules=True`.
- If you're not interested in water, I'd set `keep_solvent=False`

See also API docs: http://getyank.org/latest/api/analysis_api.html#yank.analyze.extract_trajectory

SimonBoothroyd: Thanks for the suggestions! I think we need the water for reweighting, but `image_molecules` definitely looks useful!

PR #63, opened 2019-08-02
[WIP] Revert the Use of Per-Process Tasks

SimonBoothroyd:
## Description
This PR reverts the use of per process dask tasks to allow better dask integration.

## Status
- [ ] Ready to go


SimonBoothroyd:
For now it seems like this is a useful feature to keep, and so this PR will be shelved and should be revisited and assessed at a later date.

PR #66, opened 2019-08-06
Switch to Pint Units

SimonBoothroyd:
## Description
The aim of this PR is to switch over to using `pint` units. The goal is to remove a number of awkward behaviors, including

  - a `dimensionless` quantity losing its unit after scalar multiplication.
  - handle cases such as `(x * simtk.unit.joule) / (y * simtk.unit.calorie)` being dimensionless.

## Todos
  - [x] Add conversions between `simtk.unit.X` <-> `pint.X`
  - [x] Add lots of tests for the conversions.
  - [x] Replace most instances of simtk.unit
  - [x] Run regression tests.

## Status
- [x] Ready to go

### DIFF
--- a/propertyestimator/__init__.py
+++ b/propertyestimator/__init__.py
@@ -7,6 +7,7 @@ Property calculation toolkit from the Open Forcefield Consortium.
 # Safe to remove with Python 3-only code
 from __future__ import absolute_import
 
+import os
 import pkg_resources
 
 from ._version import get_versions
@@ -26,9 +27,7 @@ for entry_point in pkg_resources.iter_entry_points('propertyestimator.plugins'):
 
 # Set up pint.
 from pint import UnitRegistry
-
 unit = UnitRegistry()
-unit.define("dalton = gram / mole = u = amu = Da")
 
 # Handle versioneer
 versions = get_versions()
--- a/propertyestimator/__init__.py
+++ b/propertyestimator/__init__.py
@@ -24,6 +24,12 @@ for entry_point in pkg_resources.iter_entry_points('propertyestimator.plugins'):
         formatted_exception = traceback.format_exception(None, e, e.__traceback__)
         logging.warning(f'Could not load the {entry_point} plugin: {formatted_exception}')
 
+# Set up pint.
+from pint import UnitRegistry
+
+unit = UnitRegistry()
+unit.define("dalton = gram / mole = u = amu = Da")
+
 # Handle versioneer
 versions = get_versions()
 __version__ = versions['version']
--- a/propertyestimator/backends/backends.py
+++ b/propertyestimator/backends/backends.py
@@ -4,7 +4,7 @@ Defines the base API for the property estimator task calculation backend.
 import re
 from enum import Enum
 
-from propertyestimator import unit as unit
+from propertyestimator import unit
 
 
 class ComputeResources:
--- a/propertyestimator/backends/backends.py
+++ b/propertyestimator/backends/backends.py
@@ -4,7 +4,7 @@ Defines the base API for the property estimator task calculation backend.
 import re
 from enum import Enum
 
-from simtk import unit
+from propertyestimator import unit as unit
 
 
 class ComputeResources:
@@ -50,9 +50,10 @@ class ComputeResources:
         self._number_of_gpus = number_of_gpus
 
         self._preferred_gpu_toolkit = preferred_gpu_toolkit
-        self._gpu_device_indices = None  # A workaround for when using a local cluster
-                                         # backend which is strictly for internal purposes
-                                         # only for now.
+        # A workaround for when using a local cluster
+        # backend which is strictly for internal purposes
+        # only for now.
+        self._gpu_device_indices = None
         
         assert self._number_of_threads >= 0
         assert self._number_of_gpus >= 0
@@ -108,7 +109,7 @@ class QueueWorkerResources(ComputeResources):
         return self._wallclock_time_limit
 
     def __init__(self, number_of_threads=1, number_of_gpus=0, preferred_gpu_toolkit=None,
-                 per_thread_memory_limit=1*(unit.giga*unit.bytes), wallclock_time_limit="01:00"):
+                 per_thread_memory_limit=1*unit.gigabytes, wallclock_time_limit="01:00"):
         """Constructs a new ComputeResources object.
 
         Notes
@@ -136,7 +137,7 @@ class QueueWorkerResources(ComputeResources):
         assert self._per_thread_memory_limit is not None
 
         assert (isinstance(self._per_thread_memory_limit, unit.Quantity) and
-                self._per_thread_memory_limit.unit.is_compatible(unit.byte))
+                unit.get_base_units(unit.byte)[-1] == unit.get_base_units(self._per_thread_memory_limit.units)[-1])
 
         assert self._per_thread_memory_limit > 0 * unit.byte
 
--- a/propertyestimator/backends/dask.py
+++ b/propertyestimator/backends/dask.py
@@ -510,7 +510,7 @@ class DaskLocalCluster(BaseDaskBackend):
 
         if self._resources_per_worker.number_of_gpus > 0:
 
-            for index, worker in enumerate(self._cluster.workers):
+            for index, worker in self._cluster.workers.items():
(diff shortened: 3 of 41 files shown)


propertyestimator/__init__.py
```diff
@@ -24,6 +25,10 @@
         formatted_exception = traceback.format_exception(None, e, e.__traceback__)
         logging.warning(f'Could not load the {entry_point} plugin: {formatted_exception}')
 
+# Set up pint.
+from pint import UnitRegistry
+unit = UnitRegistry()
```
j-wags: This is very clever, but I worry that it'll be really confusing for a contributor some day.

propertyestimator/backends/backends.py
```diff
@@ -108,7 +109,7 @@ def wallclock_time_limit(self):
         return self._wallclock_time_limit
 
     def __init__(self, number_of_threads=1, number_of_gpus=0, preferred_gpu_toolkit=None,
-                 per_thread_memory_limit=1*(unit.giga*unit.bytes), wallclock_time_limit="01:00"):
+                 per_thread_memory_limit=1*unit.gigabytes, wallclock_time_limit="01:00"):
```
j-wags: Now _that_ is really cool.

propertyestimator/backends/backends.py
```diff
@@ -4,7 +4,7 @@
 import re
 from enum import Enum
 
-from simtk import unit
+from propertyestimator import unit as unit
```
j-wags: In most of the other files, this doesn't have the `as unit` part at the end. Does that make a difference?

propertyestimator/backends/dask.py
```diff
@@ -385,7 +384,7 @@ def start(self):
         from dask_jobqueue.lsf import lsf_detect_units, lsf_format_bytes_ceil
 
         requested_memory = self._resources_per_worker.per_thread_memory_limit
-        memory_bytes = requested_memory.value_in_unit(unit.byte)
+        memory_bytes = requested_memory.to(unit.byte).magnitude
```
j-wags: I like pint already.

propertyestimator/protocols/simulation.py
```diff
@@ -10,12 +10,14 @@
...
 import yaml
-from simtk import unit, openmm
+from simtk import openmm, unit as simtk_unit
 from simtk.openmm import app
 
+from propertyestimator import unit
 from propertyestimator.thermodynamics import ThermodynamicState, Ensemble
 from propertyestimator.utils.exceptions import PropertyEstimatorException
-from propertyestimator.utils.openmm import setup_platform_with_resources
+from propertyestimator.utils.openmm import setup_platform_with_resources, openmm_quantity_to_pint,\
```
j-wags: I think it's becoming preferred to do multi-line imports using parentheses: https://www.python.org/dev/peps/pep-0328/

propertyestimator/protocols/simulation.py
```diff
@@ -394,33 +401,36 @@ def _write_statistics_array(self, raw_statistics, temperature, pressure,
...
         # systems degrees of freedom.
-        temperatures = 2.0 * kinetic_energies / (degrees_of_freedom * unit.MOLAR_GAS_CONSTANT_R)
+        temperatures = 2.0 * kinetic_energies / (degrees_of_freedom * unit.molar_gas_constant)
 
         # Calculate the systems enthalpy and reduced potential.
         total_energies = potential_energies + kinetic_energies
         enthalpies = None
 
-        reduced_potentials = potential_energies / unit.AVOGADRO_CONSTANT_NA
+        reduced_potentials = potential_energies / unit.avogadro_number
```
j-wags: (Not a suggestion for this PR, but maybe a discussion we should have about standardizing design)
I'm worried that we're going to reach a reckoning some day with the fact that we use `kJ/mole` to mean two completely separate things:
1) A very small amount of energy, like "a microfemtojoule (or whatever the proper prefix is)", that's useful for describing amounts of energy like those that a molecule encounters. 
2) An actual unit of energy divided by a unit of quantity, like "this dynamite has 100 kJ/mol of energy. If you add more moles, you get more energy."

propertyestimator/tests/test_utils/test_openmm.py
```diff
@@ -0,0 +1,191 @@
...
+    openmm_raw_value = openmm_quantity.value_in_unit(simtk_unit.gram / simtk_unit.mole)
+
+    pint_quantity = openmm_quantity_to_pint(openmm_quantity)
+    pint_raw_value = pint_quantity.to(unit.gram / unit.mole).magnitude
+
+    assert np.allclose(openmm_raw_value, pint_raw_value)
+
+
+@pytest.mark.parametrize("openmm_unit", _get_all_openmm_units())
+@pytest.mark.parametrize("value", [random(), randint(1, 10), [random(), random()], np.array([random(), random()])])
```
j-wags: This is interesting. If this is going to fail for particular values of `value`, does it make sense that we should see that failure during testing at a random time in the future (probably during unrelated development)? I think it would be better to have a --runslow option that tests a pretty exhaustive set of values, but during normal testing just have this regularly run with a fixed value. [1]

[1] The fixed value could still be random :-P https://www.xkcd.com/221/

propertyestimator/tests/test_utils/test_openmm.py
```diff
@@ -0,0 +1,191 @@
...
+
+    pint_quantity = openmm_quantity_to_pint(openmm_quantity)
+    pint_raw_value = pint_quantity.magnitude
+
+    assert np.allclose(openmm_raw_value, pint_raw_value)
+
+
+@pytest.mark.parametrize("pint_unit", _get_all_pint_units())
+@pytest.mark.parametrize("value", [random(), randint(1, 10), [random(), random()], np.array([random(), random()])])
+def test_pint_to_openmm(pint_unit, value):
```
j-wags: These are really cool tests.

propertyestimator/tests/test_utils/test_openmm.py
```diff
@@ -0,0 +1,191 @@
...
+
+            pint_quantity = openmm_quantity_to_pint(openmm_quantity)
+            pint_raw_value = pint_quantity.magnitude
+
+            assert np.isclose(openmm_raw_value, pint_raw_value)
+
+
+@pytest.mark.parametrize("openmm_unit", {*_get_all_openmm_units(), simtk_unit.dalton**2, simtk_unit.dalton**3})
+@pytest.mark.parametrize("value", [random(), randint(1, 10), [random(), random()], np.array([random(), random()])])
+def test_openmm_unit_conversions(openmm_unit, value):
```
j-wags: 👍

propertyestimator/tests/test_utils/test_openmm.py
```diff
@@ -0,0 +1,191 @@
...
+                                            isinstance(openmm_quantity, list) or
+                                            isinstance(openmm_quantity, np.ndarray)):
+        openmm_raw_value = openmm_quantity
+    else:
+        openmm_raw_value = openmm_quantity.value_in_unit(openmm_unit)
+
+    assert np.allclose(openmm_raw_value, pint_raw_value)
+
+
+def test_constants():
```
j-wags: I _really_ like how well thought-out these tests are.

propertyestimator/__init__.py
SimonBoothroyd: Unfortunately this is just the pint way of doing things: https://pint.readthedocs.io/en/0.9/tutorial.html#using-pint-in-your-projects

propertyestimator/backends/backends.py
SimonBoothroyd: Typo - thanks!

propertyestimator/protocols/simulation.py
SimonBoothroyd: Agree - will keep this for now as it's consistent, and open a PR to fix this across the codebase in the future

propertyestimator/tests/test_utils/test_openmm.py
SimonBoothroyd: This should _never_ fail in practice. If it does - it means we've found a really weird edge case that would be good to know about and `random` allows us to be on the hunt for this (even though I think it will never find any issues)

PR #99, opened 2019-09-04
Initial TLEAP Support

SimonBoothroyd:
## Description
This PR aims to add initial support for parameterising molecules with the GAFF force field using the TLEAP server.

## Status
- [x] Ready to go


propertyestimator/protocols/forcefield.py
```diff
@@ -635,3 +638,236 @@ def execute(self, directory, available_resources):
...
+    def force_field_path(self, value):
+        """The file path to the force field parameters to assign to the system.
+        This path **must** point to a json serialized `TLeapForceFieldSource` object."""
+        pass
+
+    def __init__(self, protocol_id):
+        """Constructs a new `BuildTLeapSystem` object.
+        """
+        super().__init__(protocol_id)
+
```
j-wags: Jeff will add `create_mol2_from_topology_molecule` here

j-wags: The toolkit should assign the charges (with a choice between OETK and RDK/AmberTools) before writing out the mol2.

We should generate a conformer before writing out the topology molecule as well.

The topology will be created from a PDB file, so preserving the indexing is important.

j-wags: Added prototype of this functionality in https://github.com/openforcefield/propertyestimator/pull/99/commits/05f95e946a647a9b66228f96c81e9b1d12bb70c0

j-wags:
@SimonBoothroyd -- I've added a prototype of the functionality in https://github.com/openforcefield/propertyestimator/pull/99/commits/05f95e946a647a9b66228f96c81e9b1d12bb70c0

It's probably not super stable past the next few months, so one day these unit tests will break and I'll put in something better. 

I dropped the unit test in an arbitrary location, and same with the test file. I was going to park it in a PR until we decided how to arrange things, but things have a bad habit of getting parked in PRs, so... here it is. Let me know if this doesn't do what you need it to!

propertyestimator/protocols/forcefield.py
SimonBoothroyd: Thanks @j-wags, this looks great!

PR #110, opened 2019-09-12
Protocol Cleanup and Refactor

SimonBoothroyd:
## Description
This PR aims to do a general cleanup of the workflow protocols. The main focus will be on making the input / output descriptors more general and extensible.

## TODOS

- [X] Replace the old stub functions with cleaner descriptors
- [X] Remove the need to manually define the private descriptor attribute and value.
- [X] Add automated descriptor docstrings ([e.g](https://property-estimator.readthedocs.io/en/protocol_cleanup/api/generated/propertyestimator.protocols.coordinates.BuildCoordinatesPackmol.html#propertyestimator.protocols.coordinates.BuildCoordinatesPackmol.mass_density))
- [X] Refactor merge behaviour enum to allow future extension.
- [X] Add a new `UNDEFINED` class to distinguish between `None` and unset values.
- [x] Fix existing tests.
- [x] Add per protocol tests

## Status
- [x] Ready to go


propertyestimator/protocols/analysis.py
```diff
@@ -6,13 +6,14 @@
 from os import path
 
 import numpy as np
+import typing
```
j-wags: This may be unused

propertyestimator/protocols/coordinates.py
```diff
@@ -13,96 +13,76 @@
...
+    mass_density = protocol_input(
+        docstring='The target density of the created system.',
+        type_hint=unit.Quantity,
+        default_value=0.95 * unit.grams / unit.milliliters
+    )
+
+    box_aspect_ratio = protocol_input(
+        docstring='The aspect ratio of the simulation box.',
+        type_hint=list,
+        default_value=[1.0, 1.0, 1.0]
```
j-wags: I'm not 100% sure, but it's not clear whether an object which uses this value would use a _copy_ of this list, or _the list itself_ (in which case modifying one instance of the class could affect another)

propertyestimator/protocols/forcefield.py
```diff
@@ -43,54 +43,34 @@ class WaterModel(Enum):
...
-        self._coordinate_file_path = None
-        self._substance = None
-        self._water_model = BaseBuildSystemProtocol.WaterModel.TIP3P
-
-        # Outputs
-        self._system_path = None
+    force_field_path = protocol_input(
+        docstring='The file path to the force field parameters to assign to the system.',
+        type_hint=str,
+        default_value=UNDEFINED
```
j-wags: I like `UNDEFINED` vs. `None` distinction. This has bitten me before, and I may steal the convention.

propertyestimator/utils/serialization.py
```diff
@@ -247,7 +247,7 @@ class TypedJSONEncoder(json.JSONEncoder):
         np.float64: lambda x: {'value': float(x)},
         np.int32: lambda x: {'value': int(x)},
         np.int64: lambda x: {'value': int(x)},
-        np.ndarray: lambda x: {'value': x.tolist()},
+        np.ndarray: lambda x: {'value': x.tolist()}
     }
```
j-wags: Wow. This is great. I need to steal this for serialization in the toolkit.

propertyestimator/utils/utils.py
```diff
@@ -19,7 +19,7 @@ def find_types_with_decorator(class_type, decorator_type):
     ----------
     class_type: class
         The class to pull attributes from.
-    decorator_type: str
+    decorator_type: type
         The type of decorator to search for.
```
j-wags: In the new code, is there a difference between `type`s and `class`es? [This SO post](https://stackoverflow.com/a/4479888) seems to say that they're the same thing in python 3.

propertyestimator/workflow/decorators.py
```diff
@@ -1,136 +1,267 @@
...
+    To mark an attribute as an output:
 
-    >>> @protocol_output(value_type=str)
-    >>> def coordinate_file_path(self):
-    >>>     pass
+    >>> from propertyestimator.workflow.protocols import BaseProtocol
+    >>> from propertyestimator.workflow.decorators import protocol_output,
+    >>>
+    >>> class MyProtocol(BaseProtocol):
+    >>>
```
j-wags: Having a non-indented blank line here might actually break doctests/make this code non-functional

propertyestimator/protocols/analysis.py
SimonBoothroyd: It gets used on line 132 I think!

propertyestimator/utils/utils.py
SimonBoothroyd: I _think_ the two are the same now although I'm never sure which to use for docstrings - I have swapped the old `class` to `type` in b8a79e30730682cb2478e4d77ac2ffd4b44b4b9e to at least be consistent here!

SimonBoothroyd:
@j-wags thanks for the review! We haven't got doctests set up yet, but if you want to open a PR to do so that would be welcome!

PR #127, opened 2019-10-15
Implement Solvation Free Energies Using YANK

SimonBoothroyd:
## Description
This PR aims to add the ability to estimate solvation free energies using YANK + molecular simulations. Gradients + Reweighting will be handled by a separate PR. 

## TODOS

- [X] Implement a new `SolvationYankProtocol` protocol.
- [x] Create a new `SolvationFreeEnergy` property.
- [x] Add unit tests.
- [x] Perform simple integration tests.

## Status
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -33,7 +33,7 @@ dependencies:
   - coverage >=4.4
   - uncertainties
   - openmmtools
-  - yank
+  - yank >=0.24.1
   - pyyaml
   - requests
   - requests-mock # For testing http requests.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -296,8 +296,17 @@ Built-in Workflow Protocols
 
     RunEnergyMinimisation
     RunOpenMMSimulation
+
+**YANK Free Energies**
+
+.. currentmodule:: propertyestimator.protocols.yank
+.. autosummary::
+    :nosignatures:
+    :toctree: api/generated/
+
     BaseYankProtocol
     LigandReceptorYankProtocol
+    SolvationYankProtocol
 
 **Simulation Analysis**
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -31,11 +31,10 @@ The framework currently has built in support for estimating the following proper
 +-------------------------+--------------+------------+--------------+------------+
 || V\ :sub:`excess`       || Yes         || Yes*      || Yes         || Yes*      |
 +-------------------------+--------------+------------+--------------+------------+
-|| G\ :sub:`solvation`    || Yes*,**     || No        || No          || No        |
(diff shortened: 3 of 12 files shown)


jchodera:
@SimonBoothroyd Are you intending this feature to be used for _hydration_ free energies (gas -> water), solvation free energies (gas -> solvent), or relative solvation free energies (solvent1 -> solvent2)? 

Ideally, we could use this for all three kinds of applications. I imagine hydration will be useful only once we have polarizability, while relative solubilities / solvation / partitioning free energies could be very useful in the near term (e.g. octanol -> water).

SimonBoothroyd:
> @SimonBoothroyd Are you intending this feature to be used for hydration free energies (gas -> water), solvation free energies (gas -> solvent), or relative solvation free energies (solvent1 -> solvent2)?

@jchodera The plan is to eventually use this for all three use cases, although the way it is currently setup will only work for hydration free energies and solvation free energies, and only for case of inserting a single solute molecule (i.e. not exploring these as a function of concentration, but I hope to handle this case in a future PR as needed, or possibly Owen may take this on).

I initially did want to set this up to handle relative solvation free energies straight away but I wasn't sure how to use the YANK `solvent_dsl` option with the required multiple solvents, and haven't had time to dig into this yet / see if it's actually possible (I was expecting there to be something like a `solvent1_dsl` and a `solvent2_dsl` option). It's probable that I've missed something here however and if that's the case it would be great if you could point me in the direction of the appropriate docs / code!

jchodera:
> I wasn't sure how to use the YANK solvent_dsl option with the required multiple solvents

@davidlmobley @MehtapIsik and @bergazin have recently computed octanol-water logPs with YANK, so you might just ask them to point you to their scripts. @andrrizzi could comment on how much effort it would be for us to extend the YAML input to make this easy if you don't want to use their scripts---I bet we could find some way to make these changes easily if @andrrizzi can give us some pointers.

andrrizzi:
> I was expecting there to be something like a `solvent1_dsl` and a `solvent2_dsl` option

You're right that that would be a cleaner solution. So far we never really needed two different arguments in the YAML because specifying a single `solvent_dsl` for both solvents would work unless you're decoupling a molecule of solvent. And even in that case, you can just give a different residue name for the decoupled molecule so it's never been a priority.

For example, if you're running a octanol -> water calculation you could have something like `solvent_dsl: resname OCT or resname WAT`. YANK will consider any residue named like that as solvent, regardless of the phase. If you need to decouple a molecule of water, you'd have to name that molecule differently though (e.g. `MOL`).

SimonBoothroyd:
Ah perfect I was wondering if that would work but hadn't had time to test! Glad to hear this is already possible - thanks @andrrizzi !

I'll make the change to make this class more general then before merging!

PR #132, opened 2019-10-23
[WIP] Add Basic Support For Parallel Tempering

SimonBoothroyd:
## Description
This PR aims to add basic support for performing parallel tempering simulations, making use of the fantastic utilities in [openmmtools](openmmtools.readthedocs.io). This will in principle enable more comprehensive sampling of molecules with configurations separated by large energy barriers, such as when estimating the vacuum energies of carboxylic acids and esters.

## Status
- [ ] Ready to go

### DIFF
--- a/integration_tests/scratch/run.py
+++ b/integration_tests/scratch/run.py
@@ -1,6 +1,9 @@
 import os
 import shutil
 
+import numpy as np
+import matplotlib.pyplot as plt
+
 from propertyestimator import unit
 from propertyestimator.backends import ComputeResources
 from propertyestimator.protocols import coordinates, forcefield, simulation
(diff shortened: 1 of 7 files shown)


jchodera:
GHMC is likely a poor choice for condensed phase systems since it will lead to very long correlation times. I'd highly suggest using BAOAB for condensed phase systems.

Ideally, you would use @maxentile's thresholds tool to identify the max stable timestep at high temperature, and use that throughout.

SimonBoothroyd:
> GHMC is likely a poor choice for condensed phase systems since it will lead to very long correlation times. I'd highly suggest using BAOAB for condensed phase systems.

100% agree for the condensed phase, and BAOAB will absolutely remain the default there! 

I've seeing some strange behaviour in the gas phase however whereby the distributions sampled at each replica temperature using parallel tempering didn't match the distribution produced by running just a regular simulation, so I've been trying to look into the cause of this by varying things such as the integrator / timestep. I'll update more on this if I manage to get some illuminating results.

> Ideally, you would use @maxentile's thresholds tool to identify the max stable timestep at high temperature, and use that throughout.

Agree also that this would be a great addition, unfortunately I just haven't had the time to integrate this yet.

jchodera:
Please do file an issue in openmmtools if you notice anything odd! Ideally you could I close XML files of the system and a few lines of code to reproduce the parallel tempering simulation.

SimonBoothroyd:
Closing this as stale as the PR is now very far behind master, and the implementation flawed.

PR #139, opened 2019-11-11
Change default value for max iterations for minimization

MSchauperl:
## Description
Change the default value for the maximum number of iterations for the minimization.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


jchodera:
Are you sure you want to avoid _all_ minimization?

MSchauperl:
0 does not mean no optimization. It runs optimization until it converges (based on energy criteria). 10 steps are not enough for certain hosts in the host-guest binding protocol. I talked already with @SimonBoothroyd and @slochower about this and they both agreed that the value should be 0.

SimonBoothroyd:
Thanks for this @MSchauperl, although I think I beat you to this in #138! I've merged that PR in the `paprika_integration` branch in favor of this PR.

PR #140, opened 2019-11-14
Include equilibration timestep

MSchauperl:
## Description
Different timestep for equilibration than for production. 

## Status
-I only modified it for OpenMM right now.


propertyestimator/protocols/paprika.py
```diff
@@ -513,7 +519,7 @@ def _perform_analysis(self, directory):
                 self._results_dictionary['release']['ti-block']['sem'] * unit.kilocalorie / unit.mole,
                 self._id + "_release")
 
-        if 'ref_state_work' in self._results_dictionary:
+        if 'ref_state_work' in self._results_dictiohost_protocol.schemanary:
```
SimonBoothroyd: I think this is a typo?

MSchauperl: Yes it is a typo. Thanks for catching it.

MSchauperl:
I corrected the typo and modified the amber equilibration.

PR #186, opened 2020-02-05
Add Utility to Check for OpenEye

SimonBoothroyd:
## Description
This PR adds a new `has_openeye` utility to check if the `openeye` package is available.

## Status
- [X] Ready to go


jchodera:
@SimonBoothroyd : Are there ways we can move more functionality to the openforcefield toolkit to reduce your direct dependencies on specific cheminformatics toolkits?

PR #187, opened 2020-02-05
Expand RDKit Support

SimonBoothroyd:
## Description
This PR expands support for RDKit by:

- instructing `cmiles` to use `rdkit` for smiles standardisation
- skipping OE dependant tests.
- fixing tests against OE style smiles.

## Status
- [x] Ready to go


jchodera:
@SimonBoothroyd : Once cmiles makes it into the `openforcefield` toolkit, is there a way we can make this easier by allowing some sort of global `openforcefield` toolkit option to be set (to prefer RDKit) and then hide any toolkit-dependent effects behind the `openforcefield` APIs?

PR #233, opened 2020-04-16
Fix bug when solvating structures and print simulation speed

jeff231li:
## Description
* Fixed bug when creating systems with packmol through the `SolvateExistingStructure` class. The box aspect ratio was not set properly. [Problematic for HG systems, the solvate function creates a cubic box for the attach-pull calculations instead of a rectangular box]
* ~Increased the default tolerance for packmol to 2.4 angstrom up from 2.0 angstrom. [Sometimes simulations for HG systems with the default tolerance crashes after minimization even with a 1 fs timestep. Increasing the tolerance to 2.4 angstroms seems to alleviate this issue.]~
* The simulation speed will now be printed in `openmm_statistics.csv`. [benchmarking purposes]

### DIFF
--- a/evaluator/utils/packmol.py
+++ b/evaluator/utils/packmol.py
@@ -541,7 +541,7 @@ def pack_box(
     number_of_copies,
     structure_to_solvate=None,
     center_solute=True,
-    tolerance=2.0 * unit.angstrom,
+    tolerance=2.4 * unit.angstrom,
     box_size=None,
     mass_density=None,
     box_aspect_ratio=None,
--- a/evaluator/utils/packmol.py
+++ b/evaluator/utils/packmol.py
@@ -541,7 +541,7 @@ def pack_box(
     number_of_copies,
     structure_to_solvate=None,
     center_solute=True,
-    tolerance=2.4 * unit.angstrom,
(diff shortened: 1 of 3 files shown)


evaluator/utils/packmol.py
```diff
@@ -541,7 +541,7 @@ def pack_box(
     number_of_copies,
     structure_to_solvate=None,
     center_solute=True,
-    tolerance=2.0 * unit.angstrom,
+    tolerance=2.4 * unit.angstrom,
```
SimonBoothroyd, line 544: I think in general 2.0 is an ok default for most applications, so it would be good to change this back.

```suggestion
    tolerance=2.0 * unit.angstrom,
```

Instead I have exposed this as [an input to the corresponding workflow protocol](https://github.com/openforcefield/openff-evaluator/pull/236), so this can be varied using that.

jeff231li:
Thank you for reviewing. Good idea in exposing the tolerance variable.

evaluator/utils/packmol.py
jeff231li: Thank you, I have reverted the tolerance back to 2.0 angstrom.

PR #240, opened 2020-05-19
Add omnia/label/beta to README installation instructions

mattwthompson:
So the one-liner mirrors https://openff-evaluator.readthedocs.io/en/latest/install.html#installation-from-conda


SimonBoothroyd:
Thanks for catching this @mattwthompson! Once this is merged can you also update #221 to roll-this back so it doesn't get missed once out of beta?

PR #248, opened 2020-05-29
prints job id in worker log for PBS/LSF

jeff231li:
A PR to allow each `dask-worker` to print the Job ID to the log file when running on a PBS or LSF cluster. This information is useful for tracking down errors.


SimonBoothroyd:
Thanks @jeff231li - this looks like it will be really helpful! Once the tests pass I'll approve the PR 🙂

Issue #258, opened 2020-07-10
Issue with dask worker restarting/continuing a job


jeff231li:
I've just noticed an issue with `Evaluator` restarting/picking up a job that gave similar error to issue https://github.com/openforcefield/openff-evaluator/issues/224 about appending to an empty binary file but may be caused by something else. The log file for the worker for the failed job shows a 7:56 timestamp
```
07:55:53.738 INFO     Running window 1 out of 59
07:55:53.889 INFO     Executing npt_equilibration
07:55:54.199 INFO     Setting up an openmm platform on GPU 0
07:56:00.527 INFO     No checkpoint files were found.
07:56:30.478 INFO     Protocol failed to execute: npt_equilibration
07:56:30.526 INFO     Protocol failed to execute: 45ab11237d9b4db899cb5418c9d5a5be|host_guest_free_energy_0
07:56:38.933 INFO     Executing 1307506edaf9477882944919aa4e8f65|host_guest_free_energy_1
```
Tracing back to the previous worker that dealt with this job indicates that it quit at 6:13 because it exceeded the max wall limit.
```
06:13:49.829 INFO     Executing npt_equilibration
06:13:49.831 INFO     Setting up an openmm platform on GPU 0
06:13:51.863 INFO     No checkpoint files were found.
```
and the files in the directory are
```
-rw-r--r-- 1 jsetiadi gibbs-group 548306 Jul 10 06:13 input.pdb
-rw-r--r-- 1 jsetiadi gibbs-group      0 Jul 10 06:13 openmm_statistics.csv 
-rw-r--r-- 1 jsetiadi gibbs-group      0 Jul 10 06:13 trajectory.dcd 
```
It looks like the worker dies before the DCDReporter/StateDataReporter could write anything to file. So then the next worker that picks up this job will definitely complain about appending to an empty `trajectory.dcd`.

As a potential fix, what do you think of checking for not only if the file exists but also the file size of `trajectory.dcd`? So changing https://github.com/openforcefield/openff-evaluator/blob/789885755ce3cd49f384f909c384e32dc33262d1/openff/evaluator/protocols/openmm.py#L602-L605 to
```python
append_trajectory = os.path.isfile(self._local_trajectory_path) and os.path.getsize(self._local_trajectory_path) != 0
dcd_reporter = app.DCDReporter (
    self._local_trajectory_path, 0, append_trajectory
)
```
[...]

PR #259, opened 2020-07-15
Added utility `isfile_notempty`; addresses failure modes with openmm

dotsdl:
## Description
Provide a brief description of the PR's purpose here.

Addresses #258.

Adds an additional check for file existence in the `openmm` protocol; if an output file is empty, we treat it as if it doesn't exist.

## Todos

  - [x] Determine correct place for this logic to live; currently in `openff.evaluator.utils.utils`
  - [x] Create an MCVE that can be used for tests

## Status
- [x] Ready to go

### DIFF
--- a/openff/evaluator/utils/utils.py
+++ b/openff/evaluator/utils/utils.py
@@ -258,3 +258,7 @@ def has_openeye():
         available = False
 
     return available
+
+
+def isfile_notempty(path):
+    return os.path.isfile(path) and (os.path.getsize(path) != 0)
--- a/openff/evaluator/utils/utils.py
+++ b/openff/evaluator/utils/utils.py
@@ -281,5 +281,17 @@ def has_openeye():
     return available
 
 
-def is_file_and_not_empty(path):
-    return os.path.isfile(path) and (os.path.getsize(path) != 0)
+def is_file_and_not_empty(file_path):
+    """Checks that a file both exists at the specified ``path`` and is not empty.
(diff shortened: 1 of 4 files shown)


openff/evaluator/utils/utils.py
```diff
@@ -258,3 +258,7 @@ def has_openeye():
         available = False
 
     return available
+
+
+def isfile_notempty(path):
```
SimonBoothroyd, line 263: I think it would be good to make the name a bit more verbose:
```suggestion
def is_file_and_not_empty(path):
```

SimonBoothroyd: It would also be good to expose this in `openff.evaluator.utils` by adding an entry to the `__init__.py`.

SimonBoothroyd: Finally can you add a unit test for this in `tests.test_utilts.test_utilts`.

dotsdl: Done! Thanks @SimonBoothroyd!

dotsdl:
Ready to go! Looks like black 20 20.8b1 is being pulled into the lint workflow, and this insists on changes across the whole codebase. We should probably either pin black to <20 or make a blackified commit on master soon to clear.

SimonBoothroyd (approved):
LGTM - thanks @dotsdl! Feel free to squash merge when ready.

PR #271, opened 2020-08-24
[PAPRIKA] Fix ion concentration in in system setup

jeff231li:
This PR provides a small bug fix so that the correct ion concentrations are added to the host-guest system.


jeff231li:
Oops, made a mistake in the choice of the branch to merge to.

PR #272, opened 2020-08-24
[PAPRIKA] Fix ion concentration in system setup

jeff231li:
This PR provides a small bug fix so that the correct ion concentrations are added to the host-guest system.


SimonBoothroyd (approved):
LGTM! Thanks @jeff231li

PR #273, opened 2020-09-02
Substance.get_molecules_per_component now allows adjustable component

dotsdl:
Added an `adjustable` keyword argument to `substances.Substances.get_molecules_per_component` that allows specifying a component that can be incremented/decremented as needed at the end of the function to meet the exact `maximum_molecules` specified.

Also adds:
- `Component.Role.Salt` for ions that are not considered solute but we want to distinguish from adjustable solvent components

## Description
This addresses an issue raised by @jeff231li in which `protocols.coordinates.BuildCoordinatesPackmol._build_molecule_arrays` raises an error when `total_molecules > max_molecules`, which happens often enough with the existing implementation.

## Todos
- [ ] Add unit test exercising the new `adjustable` keyword argument
- [ ] Add an opt-in parameter to `BuildCoordinatesPackmol` for use of the `adjustable` keyword for component with role 'Solvent'

## Questions
- [ ] Should `adjustable` take in multiple identifiers and cycle through them somehow in the decrement/increment scheme?

## Status
- [ ] Ready to go

### DIFF
--- a/openff/evaluator/protocols/coordinates.py
+++ b/openff/evaluator/protocols/coordinates.py
@@ -124,9 +124,16 @@ class BuildCoordinatesPackmol(Protocol):
             molecule = Molecule.from_smiles(component.smiles)
             molecules.append(molecule)
 
+        # TODO: make this optionally enabled
+        adjustable = None
+        for component in self.substance.components:
+            if component.role.name == 'Solvent':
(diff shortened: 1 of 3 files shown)


SimonBoothroyd:
Closing in favour of #275 .

PR #281, opened 2020-10-12
Initial Pass at Physical Property Documentation

### LINKED ISSUE #16: Update documentation roadmap for physical properties
At some point, we should update the [physical property roadmap documentation](https://github.com/openforcefield/propertyestimator/blame/public_api_cleanup/docs/physicalproperties.rst#L132) since this is now rather out of date.

SimonBoothroyd:
## Description
This PR adds an initial set of documentation which outlines how each of the supported physical properties are estimated by the framework.

These changes have been [rendered on read the docs here](https://openff-evaluator.readthedocs.io/en/property-documentation/properties/properties.html).

This PR closes #16.

## Status
- [x] Ready to go

### DIFF
--- a/docs/_static/css/theme_overrides.css
+++ b/docs/_static/css/theme_overrides.css
@@ -44,6 +44,9 @@ table.clean-table.property-table td:not(:first-child) {
 ul.spaced-list li {
     margin-bottom: 10px;
 }
+ol.spaced-list li {
+    margin-bottom: 10px;
+}
 
 .green-font {
     color: mediumseagreen;
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -427,6 +427,14 @@ Built-in Workflow Protocols
     ExtractUncorrelatedTrajectoryData
     ExtractUncorrelatedStatisticsData
 
+
+.. currentmodule:: openff.evaluator.properties.dielectric
+.. autosummary::
+    :nosignatures:
+    :toctree: api/generated/
+
+    ExtractAverageDielectric
+
 **Coordinate Generation**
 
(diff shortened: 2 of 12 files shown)


mrshirts:
Are there strong reasons to estimate the density by <M/V> rather than M/<V>?  Molar volume is explicitly a pressure derivative of the free energy and adds linearly; <1/V> is not.  The results are almost the same, but the small deviations behave weirdly.   I can continue to discuss this.

mrshirts:
If the uncertainty is reported by the prediction, then it should be clarified in the documentation how the uncertainty is calculated, especially for things like mixture simulations where there is error propagation from multiple simulations.

mrshirts:
Are you skipping partial molar volume intentionally at this point in time?

SimonBoothroyd:
Thanks for your comments @mrshirts .

> Are there strong reasons to estimate the density by <M/V> rather than M/? Molar volume is explicitly a pressure derivative of the free energy and adds linearly; <1/V> is not. The results are almost the same, but the small deviations behave weirdly. I can continue to discuss this.

This is just a matter of convenience - most molecular simulation engines output the density directly so we make use of this in computing the average. And as you say this should be very similar if not the same as the other approach you mention, and so this has been fine for our purposes so far.

Just to note this PR is documenting what is currently implemented. If you think the property should be computed differently I would recommend opening an issue giving justifications and evidence as to why you think the other approach should be preferred! 🙂 

> Are you skipping partial molar volume intentionally at this point in time?

Currently partial molar volume is not a property supported by this framework and so it is not documented here. If you think this would be a good property to support I would recommend opening an issue requesting this feature and your thoughts on how it should best be computed in practice.

PR #332, opened 2020-11-12
Add dependabot

mattwthompson:
Some context here, not much to add besides pointing out that it doesn't run without the YAML file in the master branch

https://github.com/openforcefield/status/pull/20#issuecomment-726129190


SimonBoothroyd (approved):
LGTM! Thanks @mattwthompson

PR #342, opened 2021-01-16
[PAPRIKA] Host-Guest Free Energy Gradients 

jeff231li:
## Description
This PR adds support for calculating the free energy gradients for host-guest systems with the `paprika` protocol.

## Status
- [ ] Ready to go

### DIFF
--- a/openff/evaluator/datasets/taproom/taproom.py
+++ b/openff/evaluator/datasets/taproom/taproom.py
@@ -398,6 +398,8 @@ class TaproomDataSet(PhysicalPropertyDataSet):
                 "n_pull_windows": n_pull_windows,
                 "release_windows_indices": [*range(len(attach_lambdas))],
                 "release_lambdas": release_lambdas,
+                "bound_window_index": [[*range(n_pull_windows)][0]],
+                "unbound_window_index": [[*range(n_pull_windows)][-1]],
             }
         )
 
--- a/openff/evaluator/properties/binding.py
+++ b/openff/evaluator/properties/binding.py
@@ -18,6 +18,8 @@ from openff.evaluator.protocols import (
 )
 from openff.evaluator.protocols.paprika.analysis import (
     AnalyzeAPRPhase,
+    ComputeFreeEnergyGradient,
+    ComputePotentialEnergyGradient,
     ComputeReferenceWork,
     ComputeSymmetryCorrection,
 )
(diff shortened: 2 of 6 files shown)


jeff231li:
Question: I noticed that the code in `_evaluate_energies` converts the potential energy to pint units but keeps it in **kilojoule/mol**. So the gradient I'm getting for energy parameters like `LJ_epsilon` becomes **kilojoule/kilocalorie**. What is the reason for keeping the potential energy in **kilojoule/mol**?

https://github.com/openforcefield/openff-evaluator/blob/ef37ac8231e07f24dc97a3fba9a4ef03ed8651b8/openff/evaluator/protocols/openmm.py#L117-L122

openff/evaluator/datasets/taproom/taproom.py
```diff
@@ -398,6 +398,8 @@ def _build_metadata(
                 "n_pull_windows": n_pull_windows,
                 "release_windows_indices": [*range(len(attach_lambdas))],
                 "release_lambdas": release_lambdas,
+                "bound_window_index": [[*range(n_pull_windows)][0]],
```
SimonBoothroyd, line 401: Should `bound_window_index` and `unbound_window_index` be lists, or can they just be single values? I think if they can be single values this will simplify `_paprika_build_end_states_protocol` by a fair bit!

jeff231li: I set `bound_window_index` and `unbound_window_index` as lists because `template_values` from `ProtocolReplicator` requires a list as input. Are you referring to setting `template_value` and `template_index` instead? https://github.com/openforcefield/openff-evaluator/blob/e55c86b634af32c431ab8cc301ff937b8201fac4/openff/evaluator/workflow/schemas.py#L136-L138

dotsdl:
Thanks @jeff231li! I've added this to my queue to review within the next week or so. Please ping if you require it more urgently!

dotsdl:
@SimonBoothroyd I'm reviewing this PR this week; are there any areas of concern preventing merge? I'll prioritize getting the CI to pass first.

SimonBoothroyd:
> @SimonBoothroyd I'm reviewing this PR this week; are there any areas of concern preventing merge? I'll prioritize getting the CI to pass first.

I think the implementation looks like a great first pass, but currently contains large areas where code either duplicates already present functionalities (`ComputePotentialEnergyGradient`) or which can be significantly reduced (the schema definitions) which will take some careful thinking through.

Unfortunately I don't have the time at the moment to look over this, and suggest the changes which will be required to be merge, so it may be some time before this will be able to be merged - my sincerest apologies for this delay!

PR #349, opened 2021-03-23
Delete working files after batch completes

SimonBoothroyd:
## Description

This PR changes the server to now by default delete the working files produced while a calculation layer is estimating a particular batch of properties.

## Status
- [X] Ready to go


SimonBoothroyd:
(cc @ocmadin)

PR #362, opened 2021-08-17
Dask slurm backend

pavankum:
## Description
Thank you @SimonBoothroyd for pointing in this direction. I added slurm support for dask jobqueue. Replicated DaskPBSBackend() and checked the output of the job script generated using the following code
```
import openff.evaluator
openff.evaluator.__file__

from openff.evaluator import unit
from openff.evaluator.backends.dask import DaskSLURMBackend
from openff.evaluator.backends import QueueWorkerResources

worker_resources = QueueWorkerResources(
    number_of_threads=1,
    number_of_gpus=1,
    preferred_gpu_toolkit=QueueWorkerResources.GPUToolkit.CUDA,
    per_thread_memory_limit= 1 * unit.gigabytes,
    wallclock_time_limit="05:59",
)

# Create the backend object.
setup_script_commands = [
    f"conda activate evaluator",
    f"module load cuda/10.1.234",
]

calculation_backend = DaskSLURMBackend(
    minimum_number_of_workers=1,
    maximum_number_of_workers=2,
    resources_per_worker=worker_resources,
    queue_name="free-gpu",
    setup_script_commands=setup_script_commands,
    extra_script_options=["--account=account-name", "--gres=gpu:V100:4"],
)

calculation_backend.start()
print(calculation_backend.job_script())
```

## Status
- [x] Ready to go

### DIFF
--- a/openff/evaluator/backends/dask.py
+++ b/openff/evaluator/backends/dask.py
@@ -101,6 +101,11 @@ class _Multiprocessor:
                             worker_logger.info(
                                 f"LSBJOBID: {os.environ.get('LSB_JOBID')}"
                             )
+                        elif os.environ.get("SLURM_JOBID") is not None:
+                            worker_logger.info(
+                                f"SLURMJOBID: {os.environ.get('SLURM_JOBID')}"
+                            )
+
                         worker_logger.info(f"PLATFORM: {platform.platform()}")
                         worker_logger.info("-----------------------------------------")
                         worker_logger.info(
@@ -714,15 +719,12 @@ class DaskSLURMBackend(BaseDaskJobQueueBackend):
         extra_script_options=None,
         adaptive_interval="10000ms",
         disable_nanny_process=False,
-        resource_line=None,
         adaptive_class=None,
     ):
 
         """Constructs a new DaskSLURMBackend object
         Parameters
         ----------
-        resource_line: str
-            The string to pass to the `#SBATCH ` line.
         Examples
         --------
         To create a SLURM queueing compute backend which will attempt to spin up
@@ -766,12 +768,10 @@ class DaskSLURMBackend(BaseDaskJobQueueBackend):
             adaptive_class=adaptive_class,
         )
 
-        self._resource_line = resource_line
 
     def _get_extra_cluster_kwargs(self):
 
         extra_kwargs = super(DaskSLURMBackend, self)._get_extra_cluster_kwargs()
-        extra_kwargs.update({"resource_spec": self._resource_line})
 
         return extra_kwargs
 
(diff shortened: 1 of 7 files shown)


pavankum:
By mistake the tutorial updates are also mixed in this update, I can revert them back or can keep them. They're cosmetic changes in setting up the colab environment using conda-colab.

openff/evaluator/backends/dask.py
```diff
@@ -101,6 +101,11 @@ def _wrapper(func, queue, args, kwargs):
                             worker_logger.info(
                                 f"LSBJOBID: {os.environ.get('LSB_JOBID')}"
                             )
+                        elif os.environ.get("SLURM_JOBID") is not None:
```
SimonBoothroyd, line 104:
```suggestion
                        elif os.environ.get("SLURM_JOB_ID") is not None:
```
I think `SLURM_JOB_ID` is the newer variable name?

openff/evaluator/backends/dask.py
```diff
@@ -695,6 +700,87 @@ def _get_cluster_class(self):
...
+        setup_script_commands=None,
+        extra_script_options=None,
+        adaptive_interval="10000ms",
+        disable_nanny_process=False,
+        adaptive_class=None,
+    ):
+
+        """Constructs a new DaskSLURMBackend object
+        Parameters
+        ----------
```
SimonBoothroyd, line 727:
```suggestion
        """
```

openff/evaluator/backends/dask.py
```diff
@@ -695,6 +700,87 @@ def _get_cluster_class(self):
...
+            cluster_type="slurm",
+            adaptive_class=adaptive_class,
+        )
+
+
+    def _get_extra_cluster_kwargs(self):
+
+        extra_kwargs = super(DaskSLURMBackend, self)._get_extra_cluster_kwargs()
+
+        return extra_kwargs
```
SimonBoothroyd, line 776:
```suggestion
```

SimonBoothroyd: You don't need this if you aren't adding any extra kwargs 🙂

SimonBoothroyd (changes requested):
Thanks for this @pavankum ! It would be good if you could undo the changes to the notebooks to keep this PR as single purpose as possible.

Could you also please add the Slurm backend to the test in `tests/test_backends/test_dask.py` on line 27?

openff/evaluator/backends/dask.py
pavankum: I have both in the env variables on my cluster but slurm manual suggests the newer one

SimonBoothroyd:
The test failures are due to license issues when running the CI from a fork so this should be good to go! Thanks for this @pavankum

PR #366, opened 2021-08-25
Support gradients of handler attributes

SimonBoothroyd:
## Description

This PR adds support for computing gradients with respect to handler attributes such as the cutoff or the alpha and beta terms of the double exponential potential defined in `smirnoff-plugins`.

## Status
- [X] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -8,6 +8,15 @@ Releases follow the ``major.minor.micro`` scheme recommended by
 * ``minor`` increments add features but do not break API compatibility
 * ``micro`` increments represent bugfix releases or improvements in documentation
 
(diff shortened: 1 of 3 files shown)


jeff231li:
@SimonBoothroyd sorry, I'm not trying to be intrusive. Might we need to add something in `_find_relevant_gradient_keys` to check for the handler attribute as well? I think right now if `parameter_key.smirks` is set to `None` then Evaluator will skip the protocol to calculate the gradient for this `parameter_key`. https://github.com/openforcefield/openff-evaluator/blob/976cf1e5b7cdd4d4fd84fa961236afd112d9b072/openff/evaluator/workflow/workflow.py#L603-L609

PR #385, opened 2021-10-04
Smirnoff -plugin fixes

jthorton:
## Description
This PR adds some minor changes to work with smirnoff-plugin modified forcefields. 

## Todos
  - [x] load plugins by default

## Questions
- [x] should the `disable_pbc` function be used to consistently modify the openmm system as `_execute` re-impliments this logic.

## Status
- [x] Ready to go


SimonBoothroyd (approved):
Ah brilliant thanks for putting these fixes in - LGTM!

PR #399, opened 2021-11-08
Use units registry provided by `openff.units`

### LINKED ISSUE #398: Switch to use of `openff-units` for `pint` registry
In order to allow use of the same unit registry throughout the `openff` stack, we should switch to use of [openff-units](https://github.com/openforcefield/openff-units) for the units registry in openff-evaluator.

mattwthompson:
Resolves #398 

I've split this up into two steps: 515f6df53f8e47e71913dea46deccaf57576a7bc is the meat of the change and 512482579e06d69799aeb7d9d8dca5296fce6518 attempts to remove all `from openff.evaluator import unit`; while not strictly necessary, it'd need to happen eventually.


openff/evaluator/__init__.py
```diff
@@ -3,11 +3,10 @@
 A physical property evaluation toolkit from the Open Forcefield Consortium.
 """
 
+from openff.units import unit
+
```
mattwthompson, line 7: This arguably should be dropped altogether, though it would break some amount of backwards compatibility.

devtools/conda-envs/test_env.yaml
```diff
@@ -40,3 +41,5 @@ dependencies:
   - taproom
   - dataclasses
   - smirnoff-plugins # For testing custom nonbonded forces
+  - pip:
+    - git+https://github.com/openforcefield/openff-units.git@openmm-dalton
```
mattwthompson, line 45: This is https://github.com/openforcefield/openff-units/pull/11; my plan is as follows:

1. Get this passing and provisionally approved
2. Merge that PR
3. Make a new release of `openff-units`
4. Update pins here as needed
5. Aim to merge this PR

mattwthompson:
Could I get another review @SimonBoothroyd @jeff231li? At least to get through step 1 [above](https://github.com/openforcefield/openff-evaluator/pull/399#discussion_r746034311) so I don't need to go through 2+ releases of `openff-units`.

SimonBoothroyd (changes requested):
LGTM once the CI is switched back to a conda package after a new `openff-units` release

mattwthompson:
Looks like everything is passing with `openff-units` v0.1.4!

SimonBoothroyd (approved):
LGTM - thanks @mattwthompson

lilyminium:
FYI, ForceBalance imports `openmm_quantity_to_pint` so a similar change might be needed there for the next release :-)

Issue #396, opened 2021-11-03
ThermoML interface likely broken


mattwthompson:
Just to note this is masking a larger issue whereby the way ThermoML serves files has now changed and probably needs seem deeper inspection / fixes.

_Originally posted by @SimonBoothroyd in https://github.com/openforcefield/openff-evaluator/issues/394#issuecomment-959815971_

jchodera:
Oh noooo.

We've had this happen before, and I thought we had impressed upon NIST TRC the importance of engaging users through a gradual community process about major changes.

@mrshirts Can you put us in touch to sort out what can be done here?

mrshirts:
I'd be happy to - @mattwthompson or @SimonBoothroyd could you write up a sentence or two with the exact details for me to send to them so we can figure this out? In the meantime, I think we're mostly using a local copy, correct?

mattwthompson:
Sorry, I only have a surface-level knowledge here. Simon (or John, or somebody else who has used it before) would be better suited to provide direction.

ocmadin:
@mrshirts The issue seems to be that our entry point where we access/download the ThermoML tarballs has been removed and changed.  It used to be an individual `.tgz` for each of the journals, for example: https://trc.nist.gov/ThermoML/JCED.tgz . Now there is a single tarball at a different URL (https://data.nist.gov/od/ds/mds2-2422/ThermoML.v2020-09-30.tgz, landing page: https://data.nist.gov/od/id/mds2-2422).  I'm not sure if any of the data has changed (my assumption would be no), but looking at the landing page it looks like they added .json files so it's possible there were other changes.

Here's an example traceback of how it's failing: 
```
Traceback (most recent call last):
  File "/home/owenmadin/Documents/python/binary-mixture-publication/data-set-curation/curate_boron_phosphorus_silicon_data.py", line 614, in <module>
    main()
  File "/home/owenmadin/Documents/python/binary-mixture-publication/data-set-curation/curate_boron_phosphorus_silicon_data.py", line 609, in main
    initial_data = prepare_initial_data()
  File "/home/owenmadin/Documents/python/binary-mixture-publication/data-set-curation/curate_boron_phosphorus_silicon_data.py", line 57, in prepare_initial_data
    initial_data = CurationWorkflow.apply(
  File "/home/owenmadin/anaconda3/envs/binary-mixture-publication/lib/python3.9/site-packages/openff/evaluator/datasets/curation/workflow.py", line 112, in apply
[...]

PR #400, opened 2021-11-09
Access ThermoML from single archive

mattwthompson:
A naive stab at fixing #396 - everything passes without skips, though I suspect I'm not finding the right tests, something was cached locally, and/or I'm missing something more obvious.

### DIFF
--- a/openff/evaluator/datasets/curation/components/thermoml.py
+++ b/openff/evaluator/datasets/curation/components/thermoml.py
@@ -60,7 +60,7 @@ class ImportThermoMLData(CurationComponent):
     def _download_data(cls, schema: ImportThermoMLDataSchema):
 
         v2020_tarball = "https://data.nist.gov/od/ds/mds2-2422/ThermoML.v2020-09-30.tgz"
-        request = requests.get(v2020, stream=True)
+        request = requests.get(v2020_tarball, stream=True)
 
         # Make sure the request went ok.
         try:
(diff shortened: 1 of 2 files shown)


openff/evaluator/datasets/curation/components/thermoml.py
```diff
@@ -59,23 +59,19 @@ class ImportThermoMLData(CurationComponent):
     @classmethod
     def _download_data(cls, schema: ImportThermoMLDataSchema):
 
-        for journal in schema.journal_names:
+        v2020_tarball = "https://data.nist.gov/od/ds/mds2-2422/ThermoML.v2020-09-30.tgz"
```
SimonBoothroyd, line 62: The 2020 tarball is actually a tarball of the journal tarballs IIRC, so might need to then further extract these.

ocmadin: I think another (lower priority issue) is that our link will periodically break if they upload new dated tarballs and remove the old ones.  Not sure the best way to avoid that but most likely would involve @mrshirts coordinating with NIST people.

ocmadin:
I think another (lower priority issue) is that our link will periodically break if they upload new dated tarballs (e.g. `ThermoMLv2021-09-30.tgz` and remove the old ones.  Not sure the best way to avoid that but most likely would involve @mrshirts coordinating with NIST people.

SimonBoothroyd:
Superseded by #402

PR #419, opened 2022-02-01
fix wallclock time format

### LINKED ISSUE #418: Wrong time format for slurm backend
the wallclock time defined [here](https://github.com/openforcefield/openff-evaluator/blob/18962755f7663c37430d2baffb6c9bd80bd0556c/openff/evaluator/backends/backends.py#L118-L121
) is wrong for slurm jobs, slurm takes it as `Minutes:Seconds`  instead of `Hours:Minutes`

pavankum:
## Description
Fixing time format to `HH:MM:SS` since slurm jobs are reading it as `MM:SS` instead of `HH:MM` as specified in current documentation. Resolves #418, job cancellations due to time limit, needed for longer duration jobs. 

## Status
- [x] Ready to go


openff/evaluator/backends/backends.py
```diff
@@ -143,7 +143,7 @@ def __init__(
             The maximum amount of memory available to each thread.
         wallclock_time_limit: str
             The maximum amount of wall clock time that a worker can run for. This should
-            be a string of the form `HH:MM` where HH is the number of hours and MM the number
+            be a string of the form `HH:MM:SS` where HH is the number of hours and MM the number
```
jeff231li, line 146: Not that important but I think you forgot to add the description for SS.

openff/evaluator/backends/backends.py
```diff
@@ -165,7 +165,7 @@ def __init__(
         assert wallclock_time_limit is not None
         assert isinstance(wallclock_time_limit, str) and len(wallclock_time_limit) > 0
 
-        wallclock_pattern = re.compile(r"\d\d:\d\d")
+        wallclock_pattern = re.compile(r"\d\d:\d\d:\d\d")
```
jeff231li, line 168: Good catch!

openff/evaluator/backends/backends.py
pavankum: Thank you for the review! Updated the description.

SimonBoothroyd:
I think this change needs to be made in the [slurm backend](https://github.com/openforcefield/openff-evaluator/blob/18962755f7663c37430d2baffb6c9bd80bd0556c/openff/evaluator/backends/dask.py#L703) itself, as other clusters such as LSF will only expect HH:MM IIRC.

pavankum:
not sure if it is pythonic way to do this but I modified slurmbackend to append `:00` to current `wallclock_time_limit` property of `resources_per_worker`

openff/evaluator/backends/dask.py
```diff
@@ -746,13 +746,13 @@ def __init__(
...
         >>>
-        >>> pbs_backend = DaskPBSBackend(minimum_number_of_workers=1,
+        >>> slurm_backend = DaskSLURMBackend(minimum_number_of_workers=1,
         >>>                              maximum_number_of_workers=10,
         >>>                              resources_per_worker=resources,
         >>>                              queue_name='gpuqueue',
         >>>                              setup_script_commands=setup_script_commands)
         """
-
+        resources_per_worker._wallclock_time_limit = resources_per_worker.wallclock_time_limit + ":00"
```
jeff231li, line 755: What if we add an `if` statement here to check if the user has already specified the wallclock in `HH:MM:SS` format and only add `:00` if they have not. What do you think?

pavankum: I think this way would serve both purposes of keeping current functionality for LSF, and also not changing the prior definition of evaluator's [documented walltime variable](https://openff-evaluator.readthedocs.io/en/stable/api/generated/openff.evaluator.backends.QueueWorkerResources.html?highlight=wallclock_time_limit#openff.evaluator.backends.QueueWorkerResources.wallclock_time_limit). 
I actually went by [this guide from slurm](https://slurm.schedmd.com/rosetta.pdf) and changed the `wallclock_time_limit` variable in my first commit but it looks like from [IBM's](https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=o-w-1) and [Lilac's](https://hpcdocs.mskcc.org/display/CLUS/LSF+bsub+flags) documentation `HH:MM:SS` is invalid for LSF jobs. So it's better if we restrict user input to `HH:MM` and convert in the backend for slurm.

jeff231li: Interesting, it looks like I haven't been following guidelines in the documentation at all :((. Looking at my scripts, I have been specifying the wallclock in `HH:MM:SS` format when running jobs in both PBS and SLURM environments (never tried LFS). Evaluator will let this pass since it is not doing an exact string matching. I guess if we really want to restrict users to use `HH:MM` format we may need to change the pattern below to `re.compile(r"^\d\d:\d\d$")`. I don't have a strong opinion on this, what do you think @pavankum?

https://github.com/openforcefield/openff-evaluator/blob/5c6f65d157627e750af76d15f55572f3918d036b/openff/evaluator/backends/backends.py#L168-L169

pavankum: Ahhh, this would be a non-issue then :anguished:. Grad student in the lab was complaining that anything out of format was throwing an error and that's why I jumped in to change it. I checked it now and it does take `HH:MM:SS` without any complaint!! 
Thank you for mentioning that! This may well be a small documentation update then "use `HH:MM` for LSF and `HH:MM:SS` for PBS/SLURM"

jeff231li: A small documentation update sounds good 👍

pavankum:
Closing this since no changes are needed other than a documentation update.

PR #423, opened 2022-02-16
Add pre-commit config, lint with Black 22.1.0

mattwthompson:
I curated a configuration from other files to match that happens in `.github/workflows/lint.yaml` and ran

```
$ git add .pre-commit-config.yaml
$ pre-commit run --all-files
```

PR #413, opened 2022-01-06
Compatibility with OpenFF Toolkit v0.11.0

mattwthompson:
## Description
This PR updates code and tests to be compatible with the [changes coming](https://openforcefield.atlassian.net/wiki/spaces/IN/pages/2217443329/Topology+Biopolymer+Refactor+Release) in a new version of the toolkit.

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] pAPRika tests
  - [x] Yank tests
  - [x] Install checkmol
  - [ ] Update docstrings

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

PR #425, opened 2022-02-21
Fix use of wrong unit package in some OpenMM tests

mattwthompson:
## Description
Tests have been failing for a month or so. I didn't look at all logs, but [last night's](https://github.com/openforcefield/openff-evaluator/runs/5267638001?check_suite_focus=true) include some failures as a result of using OpenMM units where "OpenFF" units should be used. This PR fixes the failing tests locally.

Note that the linting check will fail because of upstream changes in black. These change are incorporated into #423.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [x] Ready to go


mattwthompson:
Nevermind, this won't work until #413 is complete

PR #433, opened 2022-03-22
Refactor out the use of cmiles

### LINKED ISSUE #383: Refactor out the use of CMILES
I notice that this project is the only one in our stack that currently [(explicitly) uses CMILES](https://github.com/search?q=cmiles+org%3Aconda-forge&type=code). Given that it's no longer maintenance and all functionality (should) be in the toolkit, we should probably remove it altogether at some point.

mattwthompson:
Closes #383, see https://github.com/openforcefield/cmiles#cmiles-is-no-longer-actively-maintained for background.

One detail I noticed was `mol_to_smiles` defaults to `isomeric=True`, whereas `Molecule.to_smiles` does not.


mattwthompson:
This was brought into #413

PR #448, opened 2022-06-06
Fix arguments in DaskLocalCluster 

darelbeida:
## Description
The positional arguments used for constructing `distributed.LocalCluster` in `DaskLocalCluster` mismatch `distributed >= 2021.02.0` since the `name` argument has added into the first position in https://github.com/dask/distributed/pull/4426.

This PR fixed this problem by passing all arguments as keyword arguments. It can avoid the API change of `distributed` package in the future as far as possible.

## Status
- [x] Ready to go

### DIFF
--- a/openff/evaluator/backends/dask.py
+++ b/openff/evaluator/backends/dask.py
@@ -825,7 +825,10 @@ class DaskLocalCluster(BaseDaskBackend):
     def start(self):
 
         self._cluster = distributed.LocalCluster(
-            self._number_of_workers, 1, processes=False
+            name=None,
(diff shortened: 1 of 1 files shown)


mattwthompson:
Hi darelbeida,

Thanks for the patch! If I'm understanding things correctly, this is a backwards-compatible change even though Dask changed their API in a confusing way (a _new_ argument in the first position, displacing all other).

Some (but not all) of the tests will fail because our OpenEye license is not unpacked on forks. Don't worry about those, this is expected behavior.

PR #452, opened 2022-06-16
Add `load_plugins` argument to `SmirnoffForceFieldSource.from_path`

### LINKED ISSUE #451: Load Smirnoff plugins by default
It would be great if we could load Smirnoff force fields with plugins by default by adding the `load_plugins=True` argument [here](https://github.com/openforcefield/openff-evaluator/blob/c21faba6f3ece8ef7c19bd837e4dc5f1d0eefb69/openff/evaluator/forcefield/forcefield.py#L107). Currently I have to load the force field with the toolkit and then create the evaluator force field from that object where as we could skip this step if we allow plugins when reading the force field directly. 

Current method
```python
from openff.toolkit.typing.engines.smirnoff import ForceField
from openff.evaluator.forcefield import SmirnoffForceFieldSource

plugin_ff = ForceField('sage_with_plugins.offxml', load_plugins=True)
evaluator_ff = SmirnoffForceFieldSource.from_object(plugin_ff)

```

mattwthompson:
May resolve #451 

I created `openff/evaluator/data/test/forcefields/buckingham-force-field.offxml` by blindly following the instructions in `smirnoff-plugins`' [README](https://github.com/openforcefield/smirnoff-plugins/blob/6cc273319e18899d3dd60b0ed21b23dcc90101a5/README.md). The package is conveniently already a dependency for tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,9 +25,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macOS-latest, ubuntu-latest]
-        python-version: [3.8, 3.9]
-        openeye: [true, false]
+        os: [macOS-latest] # , ubuntu-latest]
+        python-version: [3.8], # , 3.9]
+        openeye: [true] # , false]
 
     steps:
       - uses: actions/checkout@v3.0.2
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,9 +25,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


mattwthompson:
I wanted to add a test to ensure that plugins weren't loaded without adding in the flag. That test ran locally when only calling this new unit test but always failed in CI, I think because plugins were loaded elsewhere in tests. I'm just going to drop that part of the test; if the default behavior of this method is changed in a surprising way, I'd expect some other test to be failing as well.

mattwthompson:
@jthorton could you give this a look/spin and let me know if you think it's close to what you had in mind?

PR #459, opened 2022-07-20
Organise environments

### LINKED ISSUE #457: Docs builds broken
The RTD builds are broken after #413 as a result of the new Toolkit builds requiring `python >=3.8` and the existing environment being pinned to 3.7 and/or only Sphinx versions that work with 3.7. I tried loosening the Python pin but have not had time to investigate which Sphinx pins we could update this to.

I would really like to avoid a 3.7 backdoor anywhere in our infrastructure.

cc: @Yoshanuikabundi

Yoshanuikabundi:
## Description
Closes #457. Looks like the problem was the `icu 58*` pin, which should no longer be necessary as (a) it was set years ago and the current version is 70 and (b) it is no longer present in the test environment.

Locally autosummary is taking a very long time to complete. I want to see if that happens on RTD before I spend too much energy trouble shooting it.

## Todos
  - [X] Update dependencies in rtd environment
  - [x] Make sure docs still build on RTD

## Status
- [x] Ready to go


Yoshanuikabundi:
OK I've updated the environment to match test_env.yaml and fixed the associated issues, which seem to just be to do with mocking dependencies that actually exist.

mattwthompson (approved):
LGTM and docs are building. Thanks!

CI isn't running because the actions didn't know about `master` -> `main`. I'll fix that upstream - could you then pull that change in and merge this?

devtools/conda-envs/docs_env.yaml
```diff
@@ -9,23 +9,39 @@ dependencies:
...
-    # Sphinx specific
+    # Standard dependencies
+    - openff-toolkit >=0.11.0rc3
+    - smirnoff99frosst
+    - numpy
+    - pandas
+    - openmm >=7.6
+    - networkx
+    - packmol
+    - pymbar >=3.0.5
```
j-wags, line 20: Also downpin this to under 4 - Release coming very soon

mattwthompson:
The PR build looks fine; I'll make sure pymbar stuff is working fine by the next release.

Thanks for doing the hard work here @Yoshanuikabundi - removing `icu` is definitely not among the first things I was going to try given that all-caps comment.

PR #461, opened 2022-07-21
Update for pymbar >=4

mattwthompson:


mattwthompson:
Tests passed locally via `pip` install, although the JAX that was pulled down was broken because my conda environment is under Rosetta and I didn't know how to communicate that to `pip`. This might pass on the first kick once the conda builds are up: https://github.com/conda-forge/pymbar-feedstock/pull/23

mattwthompson:
It looks like the breaking changes of `pymbar` 4.0 are only detected in tests. I still prefer pinning to 4.0+ to avoid ambiguity, but there there is a reason that the core package should be allowed to fall back to 3.x we can implement that change.

mrshirts:
Is there any thing we should be changing on the pymbar side to help with any transitions?

mattwthompson:
@mrshirts nope - this is all done and it was pretty painless for me. The migration guide was all the information I needed and it was easily searchable, which makes things very easy for me. Thanks to you and the team for putting that together!

The process was basically

1. Install the new version (I grabbed it from PyPI because I was eager to get going, but now with builds from `conda-forge` are working)
2. Run tests to see what failed
3. command + f through the migration guide to find the change (took 2 seconds)
4. Update to the new API
5. Push changes, see tests are passing, merge


There might be headaches on the packaging side if people or libraries for some reason need version 3.x, but we'll cross that bridge when we get there.

PR #463, opened 2022-07-26
Fix `None` pressure failing unit conversion

### LINKED ISSUE #450: Hvap calculations broken
I have found that when trying to repeat the evaluator example in forcebalance [here](https://github.com/leeping/forcebalance/tree/master/studies/024_openff_evaluator) the calculations fail with an error related to `openff.units` and a message in the logs about the `equilibration_simulation_gas` protocol failing to execute

```
aeec2c29c7254d44a08924d9d3cd5930|equilibration_simulation_gas failed to execute.

Traceback (most recent call last):
  File "/home/josh/mambaforge/envs/nagl/lib/python3.8/site-packages/openff/evaluator/workflow/protocols.py", line 1245, in _execute_protocol
    protocol.execute(directory, available_resources)
  File "/home/josh/mambaforge/envs/nagl/lib/python3.8/site-packages/openff/evaluator/workflow/protocols.py", line 704, in execute
  File "/home/josh/mambaforge/envs/nagl/lib/python3.8/site-packages/openff/evaluator/protocols/openmm.py", line 519, in _execute
    openmm_pressure = to_openmm(pressure)
  File "/home/josh/mambaforge/envs/nagl/lib/python3.8/site-packages/openff/utilities/utilities.py", line 75, in wrapper
    return function(*args, **kwargs)
  File "/home/josh/mambaforge/envs/nagl/lib/python3.8/site-packages/openff/units/openmm.py", line 132, in to_openmm
    value = quantity.m
AttributeError: 'NoneType' object has no attribute 'm'
```
I think as the Hvap calculations require a  gas phase simulation which sets the pressure to `None` this causes an issue when we try and convert the value to openmm compatible units [here](https://github.com/openforcefield/openff-evaluator/blob/c21faba6f3ece8ef7c19bd837e4dc5f1d0eefb69/openff/evaluator/protocols/openmm.py#L515-L519).

Maybe something like this would work
```python
 openmm_pressure = None if self.ensemble == Ensemble.NVT else to_openmm(self.thermodynamic_state.pressure)
```

mattwthompson:
Fixes #450 if I can ever get everything in sync

### DIFF
--- a/openff/evaluator/protocols/openmm.py
+++ b/openff/evaluator/protocols/openmm.py
@@ -97,7 +97,10 @@ def _evaluate_energies(
     temperature = to_openmm(thermodynamic_state.temperature)
     beta = 1.0 / (openmm_unit.BOLTZMANN_CONSTANT_kB * temperature)
 
-    pressure = to_openmm(thermodynamic_state.pressure)
+    if thermodynamic_state.pressure is None:
+        pressure = None
+    else:
+        pressure = to_openmm(thermodynamic_state.pressure)
 
     for frame_index in range(trajectory.n_frames):
 
--- a/openff/evaluator/protocols/openmm.py
+++ b/openff/evaluator/protocols/openmm.py
@@ -486,11 +486,12 @@ class OpenMMSimulation(BaseSimulation):
         temperature = self.thermodynamic_state.temperature
         openmm_temperature = to_openmm(temperature)
 
-        pressure = (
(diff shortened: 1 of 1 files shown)


mattwthompson:
In principle this should solve this issue; in reality I need to test it in the appropriate study, which is requiring a few changes upstream.

mattwthompson:
Hoping #464 reproduces the failure

PR #471, opened 2022-09-15
Finalize 0.4.0 release notes

mattwthompson:


mattwthompson:
The built docs look good to me: https://openff-docs--471.org.readthedocs.build/projects/evaluator/en/471/releasehistory.html

PR #478, opened 2022-12-07
fixes and enhancements

wenyan4work:
## Description
This PR includes multiple fixes and features.

# backends.py
restrict to one gpu per worker. I believe this is the normal case since multi-gpu running hurts the performance for the tasks supported in evaluator.

# dask.py
allow multiple workers in DaskLocalCluster backend to share one gpu through cuda-mps to improve hardware utilization and computation throughput. GPUs are assigned using round-robin assignment.

# reweighting.py
fixes `computeExpectations` for pymbar 4 API

# forcefield.py
fixes TLeap/GAFF forcefield, include:
1. incorrect charges.txt format
2. incorrect indentation leading to incorrect molecule number in assign parameter step after build coordinates
3. incorrect quit command and error log parsing when running tleap
4. outdated compute_am1bcc_charges interface to openff-toolkit

### DIFF
--- a/openff/evaluator/backends/backends.py
+++ b/openff/evaluator/backends/backends.py
@@ -92,6 +92,7 @@ class ComputeResources:
 
         if self._number_of_gpus > 0:
             assert self._preferred_gpu_toolkit is not None
+            assert self._number_of_gpus == 1, f'only 1 gpu per worker is supported'
 
     def __getstate__(self):
         return {
--- a/openff/evaluator/backends/dask.py
+++ b/openff/evaluator/backends/dask.py
@@ -805,13 +805,14 @@ class DaskLocalCluster(BaseDaskBackend):
             if visible_devices is None:
                 raise ValueError("The CUDA_VISIBLE_DEVICES variable is empty.")
 
-            gpu_device_indices = visible_devices.split(",")
 
-            if len(gpu_device_indices) != number_of_workers:
-                raise ValueError(
-                    "The number of available GPUs {} must match "
-                    "the number of requested workers {}."
-                )
+            # commented out to allow multiple calculations share one gpu through mps
+            # gpu_device_indices = visible_devices.split(",")
+            # if len(gpu_device_indices) != number_of_workers:
+            #     raise ValueError(
+            #         "The number of available GPUs {} must match "
+            #         "the number of requested workers {}."
+            #     )
 
     def start(self):
 
@@ -823,16 +824,21 @@ class DaskLocalCluster(BaseDaskBackend):
         )
 
         if self._resources_per_worker.number_of_gpus > 0:
+            assert self._resources_per_worker.number_of_gpus==1 
+            # round-robin assignment of available gpus
+            visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES")
+            gpu_device_indices = [int(id) for id in visible_devices.split(",")]
+            
 
             if isinstance(self._cluster.workers, dict):
 
                 for index, worker in self._cluster.workers.items():
-                    self._gpu_device_indices_by_worker[worker.id] = str(index)
+                    self._gpu_device_indices_by_worker[worker.id] = str(gpu_device_indices[index%len(gpu_device_indices)])
 
             else:
 
                 for index, worker in enumerate(self._cluster.workers):
-                    self._gpu_device_indices_by_worker[worker.id] = str(index)
+                    self._gpu_device_indices_by_worker[worker.id] = str(gpu_device_indices[index%len(gpu_device_indices)])
 
         super(DaskLocalCluster, self).start()
 
--- a/openff/evaluator/backends/dask.py
+++ b/openff/evaluator/backends/dask.py
@@ -805,14 +805,13 @@ class DaskLocalCluster(BaseDaskBackend):
             if visible_devices is None:
                 raise ValueError("The CUDA_VISIBLE_DEVICES variable is empty.")
 
+            gpu_device_indices = visible_devices.split(",")
 
-            # commented out to allow multiple calculations share one gpu through mps
-            # gpu_device_indices = visible_devices.split(",")
-            # if len(gpu_device_indices) != number_of_workers:
-            #     raise ValueError(
-            #         "The number of available GPUs {} must match "
-            #         "the number of requested workers {}."
-            #     )
+            if len(gpu_device_indices) != number_of_workers:
+                raise ValueError(
+                    "The number of available GPUs {} must match "
+                    "the number of requested workers {}."
+                )
 
     def start(self):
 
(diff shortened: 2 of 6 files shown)


openff/evaluator/protocols/forcefield.py
```diff
@@ -1036,7 +1042,9 @@ def _generate_charges(self, molecule):
             raise ValueError("Invalid toolkit specification.")
 
         molecule.generate_conformers(toolkit_registry=toolkit_wrapper)
-        molecule.compute_partial_charges_am1bcc(toolkit_registry=toolkit_wrapper)
+        molecule.assign_partial_charges(partial_charge_method='am1bcc',
+                                        toolkit_registry=toolkit_wrapper,
+                                        normalize_partial_charges=True)
```
mattwthompson, line 1047: Nice catch, could you apply a similar change in `openff/evaluator/protocols/coordinates.py`?

```diff
diff --git a/openff/evaluator/protocols/coordinates.py b/openff/evaluator/protocols/coordinates.py
index 59d35aa..a7055ca 100644
--- a/openff/evaluator/protocols/coordinates.py
+++ b/openff/evaluator/protocols/coordinates.py
@@ -438,7 +438,7 @@ class BuildDockedCoordinates(Protocol):
         # Assign AM1-BCC charges to the ligand just as an initial guess
         # for docking. In future, we may want to get the charge model
         # directly from the force field.
-        ligand.compute_partial_charges_am1bcc()
+        ligand.assign_partial_charges(partial_charge_method="am1bcc")
 
         return ligand.to_openeye()
 
diff --git a/openff/evaluator/protocols/forcefield.py b/openff/evaluator/protocols/forcefield.py
index 7cb3488..3f07600 100644
--- a/openff/evaluator/protocols/forcefield.py
+++ b/openff/evaluator/protocols/forcefield.py
@@ -1036,7 +1036,10 @@ class BuildTLeapSystem(TemplateBuildSystem):
             raise ValueError("Invalid toolkit specification.")
 
         molecule.generate_conformers(toolkit_registry=toolkit_wrapper)
-        molecule.compute_partial_charges_am1bcc(toolkit_registry=toolkit_wrapper)
+        molecule.assign_partial_charges(
+            partial_charge_method="am1bcc",
+            toolkit_registry=toolkit_wrapper,
+        )
 
     def _parameterize_molecule(self, molecule, force_field_source, cutoff):
         """Parameterize the specified molecule.
```
Partial charges are normalized by default from the toolkit so `normalize_partial_charges=True` is not necessary unless there's some nuance with `tleap` I'm unaware of

openff/evaluator/backends/backends.py
```diff
@@ -92,6 +92,7 @@ def __init__(
 
         if self._number_of_gpus > 0:
             assert self._preferred_gpu_toolkit is not None
+            assert self._number_of_gpus == 1, f'only 1 gpu per worker is supported'
```
mattwthompson, line 95: I'm not sure this is an improvement; even if multiple GPUs per worker is not common, I'd like to avoid locking out future use cases

openff/evaluator/backends/dask.py
```diff
@@ -823,16 +824,21 @@ def start(self):
         )
 
         if self._resources_per_worker.number_of_gpus > 0:
+            assert self._resources_per_worker.number_of_gpus==1
```
mattwthompson, line 827: Could this logic be tucked away in an `if self._resources_per_worker.number_of_gpus == 1` block?

openff/evaluator/protocols/forcefield.py
```diff
@@ -34,6 +33,7 @@
 )
 from openff.evaluator.workflow import Protocol, workflow_protocol
 from openff.evaluator.workflow.attributes import InputAttribute, OutputAttribute
+from openff.toolkit.topology import Molecule
```
mattwthompson, line 36: Since this import is just for an annotation, we can skip the import at runtime
```suggestion
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from openff.toolkit.topology import Molecule
```

openff/evaluator/protocols/forcefield.py
```diff
@@ -1004,7 +1010,7 @@ def _run_tleap(molecule, force_field_source, directory):
             os.path.join(directory, rst7_file_name),
         )
 
-    def _generate_charges(self, molecule):
+    def _generate_charges(self, molecule: Molecule):
```
mattwthompson, line 1013:
```suggestion
    def _generate_charges(self, molecule: "Molecule"):
```

openff/evaluator/protocols/reweighting.py
```diff
@@ -491,7 +491,7 @@ def _reweight_observables(
             observable_dimensions = observable.value.shape[1]
             assert observable_dimensions == 1
 
-            results = mbar.computeExpectations(
+            results = mbar.compute_expectations(
                 observable.value.T.magnitude,
```
mattwthompson, line 495: We need to support both versions for compatibility with some external packages. I have a trick for handling this that I can take care of elsewhere. I thought I did, but this must not be run in the normal test suite - any other occurrences you noticed?

openff/evaluator/backends/backends.py
jeff231li: I agree with @mattwthompson on this, there may be some use cases in the future.

openff/evaluator/protocols/reweighting.py
```diff
@@ -491,7 +491,7 @@ def _reweight_observables(
             observable_dimensions = observable.value.shape[1]
             assert observable_dimensions == 1
 
-            results = mbar.computeExpectations(
+            results = mbar.compute_expectations(
```
jeff231li, line 494: Perhaps, we can include both pymbar v3 and v4 API by using a try-except block.

mattwthompson: Yup - "my" trick for handling this is copying what I saw Pascal do a few months ago https://github.com/shirtsgroup/physical_validation/pull/220

openff/evaluator/protocols/forcefield.py
wenyan4work: tleap imposes a check of molecule (tleap calls it a residue) total net partial charge and raises warnings if the total net charge is not integral. 

and yes, `normalize_partial_charges=True` is the default. I agree it is unnecessary. I just added it to remind myself about this nuance and just in case if the default of openff-toolkit changes in the future, since the openff-toolkit API is not quite stable yet.

openff/evaluator/backends/backends.py
wenyan4work: I added this to simplify the gpu-id calculation in the round-robin gpu assignment stage.

yes, it can be removed.

openff/evaluator/protocols/forcefield.py
mattwthompson: I'm okay leaving it in if you prefer it there; it is unlikely that the toolkit will change the default to `False` but it's fine to include it to match the behavior of `tleap`

openff/evaluator/protocols/reweighting.py
mattwthompson: I'm working on this in https://github.com/openforcefield/openff-evaluator/pull/479

wenyan4work:
have to revert the changes to dask backends due to random hang problem when running with cuda-mps enabled.

PR #479, opened 2022-12-07
Improve pymbar compatibility

mattwthompson:
## Description
Provide a brief description of the PR's purpose here.

## Todos
  - [x] Look at which lines are actually run in tests
  - [x] Look for other uses of pymbar API

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/openff/evaluator/protocols/reweighting.py
+++ b/openff/evaluator/protocols/reweighting.py
@@ -22,6 +22,7 @@ from openff.evaluator.utils.observables import (
     ObservableFrame,
     bootstrap,
 )
+from openff.evaluator.utils.pymbar import _PYMBAR_MAJOR_VERSION
 from openff.evaluator.workflow import Protocol, workflow_protocol
 from openff.evaluator.workflow.attributes import InputAttribute, OutputAttribute
 
@@ -491,11 +492,19 @@ class BaseMBARProtocol(Protocol, abc.ABC):
             observable_dimensions = observable.value.shape[1]
             assert observable_dimensions == 1
 
-            results = mbar.computeExpectations(
-                observable.value.T.magnitude,
-                target_reduced_potentials.value.T.magnitude,
-                state_dependent=True,
-            )
+            if _PYMBAR_MAJOR_VERSION == 3:
+
+                results = mbar.computeExpectations(
+                    observable.value.T.magnitude,
+                    target_reduced_potentials.value.T.magnitude,
+                    state_dependent=True,
+                )
+            elif _PYMBAR_MAJOR_VERSION == 4:
+                results = mbar.compute_expectations(
+                    observable.value.T.magnitude,
+                    target_reduced_potentials.value.T.magnitude,
+                    state_dependent=True,
+                )
 
             uncertainty = results[1][-1] * observable.value.units
             average_value = average_value.plus_minus(uncertainty)
--- a/openff/evaluator/protocols/reweighting.py
+++ b/openff/evaluator/protocols/reweighting.py
@@ -22,7 +22,6 @@ from openff.evaluator.utils.observables import (
     ObservableFrame,
     bootstrap,
 )
-from openff.evaluator.utils.pymbar import _PYMBAR_MAJOR_VERSION
 from openff.evaluator.workflow import Protocol, workflow_protocol
 from openff.evaluator.workflow.attributes import InputAttribute, OutputAttribute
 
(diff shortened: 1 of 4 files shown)


openff/evaluator/protocols/reweighting.py
```diff
@@ -491,11 +492,19 @@ def _reweight_observables(
...
+                    observable.value.T.magnitude,
+                    target_reduced_potentials.value.T.magnitude,
+                    state_dependent=True,
+                )
+            elif _PYMBAR_MAJOR_VERSION == 4:
+                results = mbar.compute_expectations(
+                    observable.value.T.magnitude,
+                    target_reduced_potentials.value.T.magnitude,
+                    state_dependent=True,
+                )
```
jeff231li, line 507: Would it be possible to do the same as `detect_equilibration` and then we can just import from `openff.evaluator.utils.pymbar`?

mattwthompson: Probably, I'm just not sure about the complexities of soft importing a method vs. a function

mattwthompson: How I imagine it would work is passing method names as strings and using `getattr`; I don't imagine the methods can be passed alone since they are instance-level and not class-level

```
# pymbar.py
...

    if name == "compute_expectations":
        if _PYMBAR_MAJOR_VERSION == 3:
            return "computeExpectations"
        elif _PYMBAR_MAJOR_VERSION == 4:
            return "compute_expectations"
...

# reweighting.py
...
        from openff.evaluator.utils.pymbar import compute_expectations

            results = getattr(mbar, compute_expectations)(
                observable.value.T.magnitude,
                target_reduced_potentials.value.T.magnitude,
                state_dependent=True,
            )
...
```

Okay, the code isn't as ugly as I imagined it. The readability is okay and it successfully tucks away the logic like the other import. What do you think?

jeff231li: I think this is awesome 👍, it makes the code more compact.

openff/evaluator/protocols/reweighting.py
```diff
@@ -491,7 +491,9 @@ def _reweight_observables(
             observable_dimensions = observable.value.shape[1]
             assert observable_dimensions == 1
 
-            results = mbar.computeExpectations(
+            from openff.evaluator.utils.pymbar import compute_expectations
+
+            results = getattr(mbar, compute_expectations)(
                 observable.value.T.magnitude,
                 target_reduced_potentials.value.T.magnitude,
                 state_dependent=True,
```
mattwthompson, line 499: These lines aren't actually run in unit tests and the tests would need to be updated to feed data that converges if not bootstrapping uncertainties

mattwthompson:
It'd take me a while to understand this code enough to write a test that captures all of these changes. I'm not sure how much this particular case is hit. This at least doesn't break anything we know of, and the changes at least _look_ good in terms of the API changes. I can't find other uses in the codebase:

```shell
grep -r pymbar openff                                            12:56:16  ☁  more-pymbar ☀
openff/evaluator/tests/test_utils/test_timeseries.py:from openff.evaluator.utils.pymbar import detect_equilibration
openff/evaluator/tests/test_utils/test_timeseries.py:    """Compare the output of the ``analyze_time_series`` utility with ``pymbar``."""
openff/evaluator/utils/pymbar.py:Helper functions to simultaneously support pymbar 3 and 4.
openff/evaluator/utils/pymbar.py:    from pymbar import pymbar  # noqa
openff/evaluator/utils/pymbar.py:    import pymbar  # noqa
openff/evaluator/utils/pymbar.py:            from pymbar.timeseries import detectEquilibration as detect_equilibration
openff/evaluator/utils/pymbar.py:            from pymbar.timeseries import detect_equilibration
openff/evaluator/utils/timeseries.py:Based on the original implementation found in `pymbar
openff/evaluator/utils/timeseries.py:<https://github.com/choderalab/pymbar/tree/master/pymbar>`_
openff/evaluator/utils/timeseries.py:from pymbar.utils import ParameterError
openff/evaluator/utils/timeseries.py:    https://github.com/choderalab/pymbar. Here the code is extended support
openff/evaluator/utils/timeseries.py:    https://github.com/choderalab/pymbar. Here the code is extended support
openff/evaluator/utils/timeseries.py:            # Fix for issue https://github.com/choderalab/pymbar/issues/122
openff/evaluator/protocols/reweighting.py:import pymbar
openff/evaluator/protocols/reweighting.py:        mbar: pymbar.MBAR, target_reduced_potentials: ObservableArray
openff/evaluator/protocols/reweighting.py:        mbar = pymbar.MBAR(
openff/evaluator/protocols/reweighting.py:        mbar = pymbar.MBAR(
openff/evaluator/protocols/reweighting.py:        mbar: pymbar.MBAR,
openff/evaluator/protocols/reweighting.py:            # as pymbar does not expose an easier way to compute the average
openff/evaluator/protocols/reweighting.py:            from openff.evaluator.utils.pymbar import compute_expectations
openff/evaluator/protocols/reweighting.py:        mbar: pymbar.MBAR,
```

Personally, I'm fine with merging this as-is.

I'm not sure how to proceed here, any thoughts @jeff231li @j-wags?

PR #481, opened 2022-12-09
fix tleap parameterizing and am1bcc api

wenyan4work:
## Description
fixes TLeap/GAFF forcefield, include:

1. incorrect charges.txt format
2. incorrect indentation leading to incorrect molecule number in assign parameter step after build coordinates
3. incorrect quit command and error log parsing when running tleap
4. outdated compute_am1bcc_charges interface to openff-toolkit

fix outdated api to `assign_partial_charges(partial_charge_method="am1bcc")`

I closed the previous PR #478, and this PR includes necessary changes for the TLeap backend to work properly. 
Other modifications in #478 are either reverted or will be handled by #479 .

## Todos
None

## Questions
None

## Status
- [x] Ready to go

### DIFF
--- a/openff/evaluator/protocols/coordinates.py
+++ b/openff/evaluator/protocols/coordinates.py
@@ -438,7 +438,7 @@ class BuildDockedCoordinates(Protocol):
         # Assign AM1-BCC charges to the ligand just as an initial guess
         # for docking. In future, we may want to get the charge model
         # directly from the force field.
-        ligand.compute_partial_charges_am1bcc()
+        ligand.assign_partial_charges(partial_charge_method="am1bcc")
 
         return ligand.to_openeye()
 
--- a/openff/evaluator/protocols/forcefield.py
+++ b/openff/evaluator/protocols/forcefield.py
@@ -8,7 +8,6 @@ import logging
 import os
 import re
 import subprocess
-import textwrap
 from enum import Enum
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


mattwthompson:
Thanks for your continued patience and contributions!

The `Molecule.compute_partial_charges_am1bcc` changes are an immediate approval. I'll need to read over some code to understand the `tleap` changes; this code is [covered in some capacity](https://app.codecov.io/gh/openforcefield/openff-evaluator/blob/main/openff/evaluator/protocols/forcefield.py) by unit tests, so I'm n

This is also a good time to disclose that I'm not the original developer of this package. It takes me longer than typical to review PRs here since I often have to understand code I'm not so familiar with, but I am making an effort. If it looks like I'm figuring out some details on the fly - well, it's because I mostly am 🙂

wenyan4work:
the tleap backend requires two fixes:
1. the charges file format is not proper so when antechamber reads the partial charges it misreads this charges then throw all kinds of warnings such as total partial charge is not integral, and so on
2. the tleap backend parameterizes one molecule correctly, but the step where each molecule is added to the system, the loop body is mis-indented so only one molecule is added to the openmm system. then the following steps fail because the coordinate pdb file atom number does not match the actual atom number in the openmm system

mattwthompson:
Thanks again for the contribution, it's merged as part of #489. I was overthinking earlier and because I'm not a heavy Amber user, I wasn't sure how to start reasoning through the changes.

Issue #502, opened 2023-03-08
FilterBySmirks does not work with isotopes


jthorton:
Filtering molecules with isotopes from thermoML with evaluator does not work as the openff-toolkit molecules used during the [filtering](https://github.com/openforcefield/openff-evaluator/blob/20d206bc968e1f6fd82f161031a7b9bdd9a21f8c/openff/evaluator/datasets/curation/components/filtering.py#L818) don't track if the atom is an isotope or not. 

Example
```python
from openff.evaluator.datasets.curation.components.filtering import FilterBySmirks
from rdkit import Chem
mol_smiles = "[2H]OCCCCC"
smirks = "[2H]"
FilterBySmirks._find_smirks_matches(mol_smiles, smirks)
```
> []
```python
rdkit_mol = Chem.MolFromSmiles(mol_smiles)
rdkit_mol.GetSubstructMatch(Chem.MolFromSmarts(smirks))
```
> (0,)

Maybe it would be better to use rdkit for this function or provide a specific filter which can handle this using a simple search in the string like `mol_smiles.find("[2H]")`?

conda list openff
```# Name                    Version                   Build  Channel
openff-evaluator          0.3.11+7.g48535e0           dev_0    <develop>
openff-forcefields        2.0.0              pyh6c4a22f_0    conda-forge
openff-recharge           0.1.1              pyhd8ed1ab_0    conda-forge
openff-toolkit-base       0.10.6             pyhd8ed1ab_0    conda-forge
openff-units              0.1.5              pyh6c4a22f_0    conda-forge
openff-utilities          0.1.3              pyh6c4a22f_0    conda-forge
```

mattwthompson:
Wow - nice catch. This causes tons of problems in the toolkit, actually:

```python3
>>> [atom.mass.m for atom in Molecule.from_smiles("[2H]O[2H]").atoms]
[1.007947, 15.99943, 1.007947]
>>> Molecule.from_smiles("[13C]")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mattthompson/mambaforge/envs/smirnoff-plugins-test/lib/python3.9/site-packages/openff/toolkit/topology/molecule.py", line 1807, in from_smiles
    molecule = toolkit_registry.call(
  File "/Users/mattthompson/mambaforge/envs/smirnoff-plugins-test/lib/python3.9/site-packages/openff/toolkit/utils/toolkit_registry.py", line 356, in call
    raise e
[...]

PR #503, opened 2023-03-09
Bypass toolkit when filtering by SMIRKS

mattwthompson:
## Description
This PR implements a suggestion in #502

## Todos
  - [x] Support filtering by SMARTS/SMIRKS(?) patterns with isotopes
  - [x] Use RDKit directly in SMIRKS filtering
  - [x] Only look for one match since the result is the same if one or many matches are found

## Questions
- [x] Since we only care about whether or not there is _any_ match, do we really need to return the matches themselves and care about the atom indices?

## Status
- [x] Ready to go


openff/evaluator/datasets/curation/components/filtering.py
```diff
@@ -883,21 +883,91 @@ def _find_smirks_matches(smiles_pattern, *smirks_patterns):
...
+            map_list = [idx_map[x] for x in sorted(idx_map)]
+
+            match_kwargs = dict(
+                uniquify=False, maxMatches=2**32 - 1, useChirality=True
+            )
+
+            return [
+                tuple(match[x] for x in map_list)
+                for match in rdmol.GetSubstructMatches(qmol, **match_kwargs)
+            ]
```
mattwthompson, line 963:
```suggestion
            return len(rdmol.GetSubstructMatches(qmol, **match_kwargs)) > 0
```

I think this could just be `_has_match(...) -> bool` since the indices aren't used by the logic later.

jthorton:
Thanks @mattwthompson this looks great, I think your idea of just returning a bool for if there is a match or would be enough here and keep it nice and simple!

openff/evaluator/datasets/curation/components/filtering.py
```diff
@@ -831,21 +831,73 @@ def _find_smirks_matches(smiles_pattern, *smirks_patterns):
...
+            Taken from https://github.com/openforcefield/openff-toolkit/blob/0.12.1/openff/toolkit/utils/rdkit_wrapper.py#L2306
+            """
+            # TODO: This function does not handle SMILES parsing failures gracefully.
+            qmol = Chem.MolFromSmarts(smarts)
+
+            match_kwargs = dict(
+                uniquify=False, maxMatches=2**32 - 1, useChirality=True
+            )
+
+            return len(rdmol.GetSubstructMatches(qmol, **match_kwargs)) > 0
```
jthorton, line 893: Would using `GetSubstructMatch` work here? I think the only difference is that this method returns a single match and so breaks after finding the first instance of a match.

mattwthompson: Seems good - I don't know the RDKit API that well and didn't look here.

If you have one handy, could you share a script that uses this filter on a large dataset? You mentioned the 70k ThermoML compounds earlier. I should know my way around Evaluator better than this.

jthorton: Sure this script currently returns an empty dataset when it should contain a few records with properties of heavy water 
```python
import pandas
from openff.evaluator.datasets.curation.components import filtering, thermoml
from openff.evaluator.datasets.curation.workflow import CurationWorkflow, CurationWorkflowSchema

N_PROCESSES = 2

# download thermol and cache
thermoml_data_frame = thermoml.ImportThermoMLData.apply(
        pandas.DataFrame(),
        thermoml.ImportThermoMLDataSchema(cache_file_name="thermoml.csv"),
        N_PROCESSES,
    )

# filter the dataset to only keep heavy water
curation_schema = CurationWorkflowSchema(component_schemas=[filtering.FilterBySmirksSchema(smirks_to_include=["[2H]O[2H]"], allow_partial_inclusion=True)])

# pandas dataframe of properties
heavy_water_dataset = CurationWorkflow.apply(thermoml_data_frame, curation_schema, N_PROCESSES)

heavy_water_dataset.to_csv("heavy_water.csv")
```

mattwthompson:
There appear to be some performance impacts. Plucking a small data set and doing some basic filtering shows no change when using an inclusive SMARTS pattern and some speedup when using an exclusive one. I spent some time with the full ThermoML dataset but it's too large to run through several times.

```python
In [1]: from openff.evaluator.datasets.curation.components import filtering, thermoml
/Users/mattthompson/software/openff-evaluator/openff/evaluator/protocols/yank.py:62: YankDeprecrationWarning: The YANK protocol has been deprecated and will be removed in a future release. Free energy calculations will be enabled by Perses in the future.
  warnings.warn(

In [2]: data_set = thermoml.ThermoMLDataSet.from_doi(
   ...:     "10.1016/j.fluid.2013.10.034",
   ...:     "10.1021/je1013476",
   ...: )

In [3]: %%timeit -n 1 -r 1
   ...: filtering.FilterBySmirks.apply(
   ...:     data_set,
   ...:     filtering.FilterBySmirksSchema(
   ...:         smirks_to_include=["C#C"], allow_partial_inclusion=True
   ...:     ),
   ...: )
   ...:
   ...:
# upstream/main:
312 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
# This PR:
106 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

In [4]: %%timeit
   ...: filtering.FilterBySmirks.apply(
   ...:     data_set,
   ...:     filtering.FilterBySmirksSchema(
   ...:         smirks_to_include=["C"], allow_partial_inclusion=True
   ...:     ),
   ...: )
   ...:
   ...:
# upstream/main:
15 s ± 54.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# This PR:
15.5 s ± 175 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

A mostly-permissive pattern (`C=O`, hits ~95% of this data set) also shows a minor slowdown.

I'm not sure why this would be the case, I thought I was stripping out steps after copying the code over, not adding steps. I'm going to move forward with this, anyway, but we can revisit this later if the performance here is an issue.

PR #517, opened 2023-07-17
Add support for Foyer forcefields

aehogan:
## Description
Add support for Foyer forcefields

## Todos
  - [x] Add ForceFieldSource and build system protocol
  - [x] Add tests
  - [x] Debugging

## Questions
- [x] When running a test enthalpy of vaporization calculation I get NotImplementedError() in assign_parameters_gas

## Status
- [x] Ready to go


mattwthompson:
Wow, this is a nice surprise!

Make sure you add `foyer` into the test environment.

Don't worry about the OpenEye parts of the matrix, but you can force the tests to run in those cases by adding `if: always()` to that step in the workflow.

mattwthompson:
The traceback looks like it's not coming from a development/editable install - does `from openff.evaluator import __file__; print(__file__)` point to your clone of this repo?

openff/evaluator/forcefield/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .forcefield import (
     ForceFieldSource,
+    FoyerForceFieldSource,
```
mattwthompson, line 3: Add this to `__all__` below and the linter should be happy

openff/evaluator/forcefield/forcefield.py
```diff
@@ -261,3 +261,43 @@ def __setstate__(self, state):
...
+    """A wrapper around Foyer force fields"""
+
+    @property
+    def foyer_source(self):
+        """str: Foyer force field source."""
+        return self._foyer_source
+
+    @property
+    def cutoff(self):
+        """openff.evaluator.unit.Quantity: The non-bonded interaction cutoff."""
```
mattwthompson, line 276: Are these `openff.evaluator.unit.Quantity`s or `openff.units.unit.Quantity`s? The former is an alias of the latter, so it's not a big deal. I'm just curious what type they are presented as in this code

openff/evaluator/forcefield/forcefield.py
```diff
@@ -261,3 +261,43 @@ def __setstate__(self, state):
...
+
+    def __init__(self, foyer_source="", cutoff=0.9 * unit.nanometer):
+        """Constructs a new FoyerForceField Source
+
+        Parameters
+        ----------
+        foyer_source: str
+            'oplsaa' or a Foyer XML forcefield file
+        cutoff: openff.evaluator.unit.Quantity
+            The non-bonded interaction cutoff.
```
mattwthompson, line 287:
```suggestion
        cutoff: openff.evaluator.unit.Quantity
            The non-bonded interaction cutoff, default 0.9 nanometers.
```

openff/evaluator/protocols/forcefield.py
```diff
@@ -1052,3 +1063,54 @@ def _execute(self, directory, available_resources):
...
+        openmm.System
+            The parameterized system.
+        """
+        import mdtraj as md
+        from foyer import Forcefield
+        from openff.interchange import Interchange
+        from openff.toolkit import Topology
+
+        topology: Topology = molecule.to_topology()
+        topology.mdtop = md.Topology.from_openmm(topology.to_openmm())
```
mattwthompson, line 1093: Is this still needed? I thought I got rid of `Topology.mdtop` in recent versions (0.3.0+) of Interchange

openff/evaluator/protocols/forcefield.py
```diff
@@ -1052,3 +1063,54 @@ def _execute(self, directory, available_resources):
...
+        force_field_source: FoyerForceFieldSource
+            The foyer source which describes which parameters to apply.
+
+        Returns
+        -------
+        openmm.System
+            The parameterized system.
+        """
+        import mdtraj as md
+        from foyer import Forcefield
```
mattwthompson, line 1088: I go back and forth on whether an alias like this does more harm (longer class names) than good (making it clear which package a similarly-named class if from). What do you think?
```suggestion
        from foyer import Forcefield as FoyerForcefield
```

aehogan:
Yes, I pip installed it to test it with the production script

openff/evaluator/protocols/forcefield.py
```diff
@@ -1052,3 +1063,54 @@ def _execute(self, directory, available_resources):
...
+        topology.mdtop = md.Topology.from_openmm(topology.to_openmm())
+
+        force_field: Forcefield
+        if force_field_source.foyer_source.lower() == "oplsaa":
+            force_field = Forcefield(name="oplsaa")
+        else:
+            force_field = Forcefield(forcefield_files=force_field_source.foyer_source)
+
+        interchange = Interchange.from_foyer(topology=topology, force_field=force_field)
+        interchange["vdW"].mixing_rule = "lorentz-berthelot"
```
mattwthompson, line 1102: I don't think we should be doing this - AFAIR OPLS-AA is defined to use geometric mixing rules and we should try to respect that. I've put some work recently into making sure Interchange can handle it. Have you run into issues using it here? Needing to split `NonbondedForce` out into a bunch of custom forces is what would cause headaches, if I had to guess.

openff/evaluator/tests/test_protocols/test_forcefield.py
```diff
@@ -161,3 +163,65 @@ def download_callback(_, context):
...
+  <Type name="OW_tip4pew" class="OW_tip4pew" element="O" mass="16.0"/>
+ </AtomTypes>
+ <NonbondedForce coulomb14scale="0.5" lj14scale="0.5">
+  <Atom type="MW" charge="0.0" sigma="1.0" epsilon="0.0"/>
+  <Atom type="HW_tip4pew" charge="0.0" sigma="1.0" epsilon="0.0"/>
+  <Atom type="OW_tip4pew" charge="0.0" sigma="0.316435" epsilon="0.680946"/>
+ </NonbondedForce>
+</ForceField>"""
+            )
+
```
mattwthompson, line 216: Does this actually build a system with a virtual site? Last I was aware, Foyer didn't support that

openff/evaluator/forcefield/forcefield.py
aehogan: Good question, I copied the doc string from the other forcefield sources which use that terminology

openff/evaluator/protocols/forcefield.py
aehogan: Good point, I got this code from https://docs.openforcefield.org/projects/interchange/en/stable/using/migrating.html where LB mixing were forced onto OPLS-AA

openff/evaluator/protocols/forcefield.py
aehogan: Seems like it is not needed, I was going by the Interchange documentation https://docs.openforcefield.org/projects/interchange/en/stable/using/migrating.html

openff/evaluator/forcefield/forcefield.py
mattwthompson: Cool - it looks like all of the docstrings currently use `openff.evaluator.unit` so it's best to keep it that way for now. Consistency over accuracy, I guess

openff/evaluator/protocols/forcefield.py
mattwthompson: Those docs are probably outdated - thanks for bringing this to my attention!

openff/evaluator/protocols/forcefield.py
mattwthompson: Once again my error, and once again thanks for bringing this up!

mattwthompson:
The error I'm seeing in tests:

> openff.interchange.exceptions.UnsupportedCutoffMethodError: When using `openmm.CustomNonbondedForce`, vdW and electrostatics cutoff methods must agree on whether or not periodic boundary conditions should be used.OpenMM will throw an error. Found vdw method 1, and electrostatics method 0,

usually stems from the periodicity not being set. If `Topology.box_vectors` is set, which I assume it would be after the PACKMOL protocol finishes, then `Interchange.box` should also be set. Unless the error bubbles out from a system that is _not_ periodic, in which case you're running up against a nasty confluence of corner cases in which the correct behavior is undefined (https://github.com/openforcefield/standards/issues/7). We're hoping to resolve this within a month or so (we have a meeting in two weeks about it) but as a short-term stop-gap, you could just set the gas-phase calculations to have a large (say, 5 nm or greater) periodic box. If you choose this option please note it in a comment in the code so it can be reverted later.

openff/evaluator/tests/test_protocols/test_forcefield.py
```diff
@@ -200,17 +200,21 @@ def test_build_foyer_xml_system():
...
-  <Atom type="OW_tip4pew" charge="0.0" sigma="0.316435" epsilon="0.680946"/>
+  <Atom type="HW_tip3p" charge="0.417" sigma="1.0" epsilon="0.0"/>
+  <Atom type="OW_tip3p" charge="-0.834" sigma="0.315007" epsilon="0.63681228"/>
  </NonbondedForce>
+ <HarmonicBondForce>
+  <Bond class1="OW_tip3p" class2="HW_tip3p" length="0.09572" k="1884.06"/>
+ </HarmonicBondForce>
+ <HarmonicAngleForce>
+  <Angle class1="HW_tip3p" class2="OW_tip3p" class3="HW_tip3p" angle="1.82421813" k="230.274"/>
+ </HarmonicAngleForce>
```
mattwthompson, line 217: TIP3P is supposed to be a fixed model, but I don't think Foyer support constraints in its force field schema? We always did that at simulation runtime, I think

openff/evaluator/protocols/forcefield.py
```diff
@@ -1100,6 +1100,10 @@ def _parameterize_molecule(self, molecule, force_field_source, cutoff):
 
         interchange = Interchange.from_foyer(topology=topology, force_field=force_field)
 
+        openmm_pdb_file = app.PDBFile(self.coordinate_file_path)
+        if openmm_pdb_file.topology.getPeriodicBoxVectors() is not None:
+            interchange.box = openmm_pdb_file.topology.getPeriodicBoxVectors()
+
```
mattwthompson: Does this mean `Interchange.from_foyer` isn't parsing the box vectors on the `topology` argument? If so, I think that's a bug

aehogan: `topology` is not an argument to `_parameterize_molecule`, just the `molecule` itself which contains no box information, the better fix would be to pass the topology to `_parameterize_molecule` but I'm trying to make the smallest changes I can

aehogan: the `topology` being passed to interchange is from `molecule.to_topology()` not from `TemplateBuildSystem._execute.topology` (which isn't passed as an argument to `_parameterize_molecule`)

mattwthompson: That makes sense, I was misreading the code then

openff/evaluator/tests/test_protocols/test_forcefield.py
aehogan: The test builds a rigid tip3p model HOWEVER this seems due to a short circuit in `TemplateBuildSystem._execute` that forces all water to be rigid tip3p with `TemplateBuildSystem._build_tip3p_system`. Is the intent to always override any water with tip3p in evaluator?

mattwthompson: I don't think that's the intent

mattwthompson:
@aehogan I think this is ready to go? Is there anything more you were looking to do, and have you been able to use this in your studies? If there aren't issues you've ran into, I think I'll give this another review and look to merge it into a release soon. How does that sound to you?

The changeset is smaller than I originally anticipated, but thinking through it again, it shouldn't be a surprise that it doesn't need to do much more than swap out the force field source.

aehogan:
@mattwthompson There are still one or two more issues with running a full delta H vaporization simulation I need to iron out.

Besides that I was going to change out the xml test with a methane so that part of the code is actually tested.

mattwthompson:
Sounds good, let me know when you think it's ready - no rush

aehogan:
@mattwthompson should be good now

mattwthompson (approved):
I once again trust this is working for simulations in your research; the code looks great but it's hard to say much comprehensively without battle-testing it since Evaluator relies so much on things that take a long time to test.

Could you also merge the `main` branch back in add a one-liner to the "Current development" section? Copying the name of the PR is plenty.

mattwthompson:
Awesome work!

aehogan:
Thanks! Delta H vaporization, delta H mixing and densities have all run successfully now.

PR #527, opened 2023-09-13
Use interchange to create openmm system over openff-toolkit

aehogan:
## Description
Use openff interchange to create openmm system over the openff-toolkit's create_openmm_system

## Todos
 - [ ] Test with real world simulations

## Status
 - [ ] Ready to go


j-wags:
Two things:

* Matt's out this week and may be able to review this when he returns on September 18
* Half of the tests are failing because we need to use some secret values in the OpenEye-dependent tests, and those secret values aren't available if the PR comes from a fork. I think you should have access to make a branch of the actual repo and open a new PR from that if you want to give it a shot (or let me know on Slack and I'll show you how to do this)

PR #528, opened 2023-09-15
Use Interchange functionality to create openmm system

aehogan:
## Description
Use openff interchange to create openmm system over the openff-toolkit's create_openmm_system

## Todos
 - [ ] Test with real world simulations

## Status
 - [ ] Ready to go

### DIFF
--- a/openff/evaluator/protocols/forcefield.py
+++ b/openff/evaluator/protocols/forcefield.py
@@ -13,6 +13,7 @@ from enum import Enum
 import numpy as np
 import openmm
 import requests
+from openff.interchange import Interchange
 from openff.units import unit
 from openff.units.openmm import to_openmm
 from openmm import app
@@ -561,7 +562,16 @@ class BuildSmirnoffSystem(BaseBuildSystem):
             pdb_file.topology, unique_molecules=unique_molecules
         )
 
-        system = force_field.create_openmm_system(topology)
+        interchange = Interchange.from_smirnoff(
+            topology=topology, force_field=force_field
+        )
+        # Do not combine nonbonded forces if any forcefield collection is a plugin
+        combine_nonbonded_forces = not any(
+            [collection.is_plugin for collection in interchange.collections.values()]
+        )
+        system = interchange.to_openmm(
+            combine_nonbonded_forces=combine_nonbonded_forces
+        )
 
         if system is None:
             raise RuntimeError(
--- a/openff/evaluator/protocols/forcefield.py
+++ b/openff/evaluator/protocols/forcefield.py
@@ -565,10 +565,17 @@ class BuildSmirnoffSystem(BaseBuildSystem):
         interchange = Interchange.from_smirnoff(
             topology=topology, force_field=force_field
         )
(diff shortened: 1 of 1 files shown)


openff/evaluator/protocols/forcefield.py
```diff
@@ -561,7 +562,16 @@ def _execute(self, directory, available_resources):
...
         )
 
-        system = force_field.create_openmm_system(topology)
+        interchange = Interchange.from_smirnoff(
+            topology=topology, force_field=force_field
+        )
+        # Do not combine nonbonded forces if any forcefield collection is a plugin
+        combine_nonbonded_forces = not any(
+            [collection.is_plugin for collection in interchange.collections.values()]
+        )
```
mattwthompson, line 571: This would work for now, but would break if any future plugin has a different requirement that what is currently used. Could you use a pattern like [here](https://github.com/openforcefield/openff-interchange/blob/8ab8cf3f07d07507f31ec6ce2d2ce4e0fc94ce5c/openff/interchange/interop/openmm/__init__.py#L62-L70) to determine if the keyword is needed, and then pass it through?


```suggestion
        combine_nonbonded_forces = True

        # Loop through plugin handlers ...
        for collection in interchange.collections.values():
            try:
                ... and if this needs to be False ...
                collection.check_openmm_requirements(
                    combine_nonbonded_forces=combine_nonbonded_forces,
                )
            except AssertionError:
                ... set it to False and keep going
                combine_nonbonded_forces = False
                break
```
I haven't run this code, hopefully it's legible enough to carry the message

aehogan: That is a better idea but eventually will need a way to separate out which option is throwing the exception. I've tested this code end to end with a bunch of different forcefields and it's working now.

aehogan: @mattwthompson Do you need anything else for this?

mattwthompson: Sorry, I've been out of the country and a bit behind on my maintenance duties. Thanks for testing the code - that it's passing CI is nice but your assurances that it works in other situations is also encouraging!

openff/evaluator/protocols/forcefield.py
```diff
@@ -561,7 +562,23 @@ def _execute(self, directory, available_resources):
...
+            try:
+                collection.check_openmm_requirements(
+                    combine_nonbonded_forces=combine_nonbonded_forces
+                )
+            except AssertionError:
+                combine_nonbonded_forces = False
+
+        system = interchange.to_openmm(
+            combine_nonbonded_forces=combine_nonbonded_forces
+        )
```
mattwthompson, line 581: At some point we might want to think about how keywork arguments would be wired through here, including whether or not they should be. But that time is definitely not now

mattwthompson (approved):

openff/evaluator/protocols/forcefield.py
aehogan: Agreed.

PR #533, opened 2023-10-12
Add AmoebaMultipoleForce to disable_pbc

aehogan:
## Description
Support AmoebaMultipoleForce in utils/openmm/disable_pbc

## Todos
  - [x] Support AmoebaMultipoleForce

## Status
- [x] Ready to go

### DIFF
--- a/openff/evaluator/utils/openmm.py
+++ b/openff/evaluator/utils/openmm.py
@@ -132,12 +132,16 @@ def disable_pbc(system):
     for force_index in range(system.getNumForces()):
         force = system.getForce(force_index)
 
-        if isinstance(force, openmm.NonbondedForce):
-            force.setNonbondedMethod(openmm.NonbondedForce.NoCutoff)
-        elif isinstance(force, openmm.CustomNonbondedForce):
-            force.setNonbondedMethod(openmm.CustomNonbondedForce.NoCutoff)
-        elif isinstance(force, openmm.AmoebaMultipoleForce):
-            force.setNonbondedMethod(openmm.AmoebaMultipoleForce.NoCutoff)
+        if isinstance(
+            force,
+            (
+                openmm.NonbondedForce,
+                openmm.CustomNonbondedForce,
+                openmm.AmeobaMultipoleForce,
+            ),
+        ):
+            # NoCutoff = 0, NonbondedMethod.CutoffNonPeriodic = 1
+            force.setNonbondedMethod(0)
 
 
 def system_subset(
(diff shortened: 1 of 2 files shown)


aehogan:
@mattwthompson can you look at this whenever you have a chance?

openff/evaluator/utils/openmm.py
```diff
@@ -132,12 +132,12 @@ def disable_pbc(system):
...
-
-        force.setNonbondedMethod(
-            0
-        )  # NoCutoff = 0, NonbondedMethod.CutoffNonPeriodic = 1
+        if isinstance(force, openmm.NonbondedForce):
+            force.setNonbondedMethod(openmm.NonbondedForce.NoCutoff)
+        elif isinstance(force, openmm.CustomNonbondedForce):
+            force.setNonbondedMethod(openmm.CustomNonbondedForce.NoCutoff)
+        elif isinstance(force, openmm.AmoebaMultipoleForce):
+            force.setNonbondedMethod(openmm.AmoebaMultipoleForce.NoCutoff)
```
mattwthompson, line 140: If I understand correctly, this force's non-bonded methods mirror the same behavior in `NonbondedForce` etc. 

```python
>>> import openmm
>>> type(openmm.NonbondedForce.NoCutoff)
<class 'int'>
>>> openmm.NonbondedForce.NoCutoff == openmm.CustomNonbondedForce.NoCutoff == openmm.AmoebaMultipoleForce.NoCutoff
True
>>> openmm.AmoebaMultipoleForce.NoCutoff == 0
True
```
I think this would be a simpler solution, and it would also enable sneaking other force classes in here if that's needed in the future (are there others you'd want to include in this block?)

```suggestion
        if isinstance(
            force,
            (
                openmm.NonbondedForce,
                openmm.CustomNonbondedForce,
                openmm.AmeobaMultipoleForce,
            ),
        ):
            # NoCutoff = 0, NonbondedMethod.CutoffNonPeriodic = 1
            force.setNonbondedMethod(0)
```

Would this work in your current pipeline?

aehogan: That would work, I currently don't have any other forces to add at the moment, is trusting that openmm won't change their enums the best though?

mattwthompson: I think OpenMM changing these enums in particular would be uncharacteristic of how slow and stable the development has been for the past few years.

I suppose something like `force.setNonbondedMethod(getattr(force, "NoCutoff"))` could safeguard against that. I just don't prefer `if` blocks that do effectively the same thing in each case.

mattwthompson (approved):
LGTM - it'd be nice to have a test, but the code it's changing was already missing coverage.

Could you add a one-liner to the release notes? (I think I forgot to add a section for 0.4.6 ... I'll clean that up later.)

Issue #486, opened 2022-12-15
Virtual sites missing from gradient calculations


jthorton:
I think that virtual sites are currently missing from property gradient calculations but should be included if present. Currently, evaluator tries to make a system with a reduced number of force field parameters when calculating the gradients of properties [here](https://github.com/openforcefield/openff-evaluator/blob/7230c0fa431e4d7d51c080a4aca7ac3fb1103773/openff/evaluator/utils/openmm.py#L149) and tries to make sure that any interdependencies between parameters are accounted for but virtual sites seem to be missed. I think they should be included and it would be great if not just that handler is included but also any subclasses of it are as well, this is related to the custom vsite handler we currently use in QUBEKit and this [fix](https://github.com/openforcefield/openff-toolkit/pull/1470) in the toolkit might be relavent here.

mattwthompson:
Could you write up a minimal reproduction of this? My guess is that Interchange is failing to create some of the virtual sites, but it's also possible that something is misfiring with the gradient keys.

mattwthompson:
If I understand your suggestion correctly, is the idea is add this logic to `system_subset`? Or should Interchanges behavior around this case be change?

jthorton:
I think the logic should be in `system_subset` but will defer to you here. 

I think the main difficulty is finding all the related parameters to the target. For example, fitting the v-site properties in mixtures will require the AM1BCC charge handler if that was used to parameterise the other molecules and when plugin handlers are involved it becomes more complicated again, how would we ensure we have included all present plugins which are related?

mattwthompson:
I think I misunderstood the problem originally; the error bubbles up from Interchange's rule about needing an `ElectrostaticsHandler` when _anything_ might get a charge, so either Evaluator would need to always provide that (and probably other handlers that assign methods) or Interchange's rule about this behavior would need to change.

Does the gradient-fitting routine work when a "subset" still contains so many handlers? Here's where my lack of actual knowledge of the code and science limits me.

jthorton:
> so either Evaluator would need to always provide that (and probably other handlers that assign methods) or Interchange's rule about this behavior would need to change.

Evaluator would need to provide all present handlers which control the same interaction as the one we want the gradient of including plugins. I can see two possible ways we could do this:
- 1. only use the system subset when there are no plugins present. We can hard code the relationships between the parameters so include all used electrostatic handlers when the gradient is requested for one of them and include every handler when a plugin is present in the system.
[...]

PR #534, opened 2023-10-12
Include `VirtualSiteHandler` and dependencies in gradient subset

mattwthompson:
## Description
This PR is a minimal change to get [this](https://github.com/openforcefield/openff-evaluator/issues/486#issuecomment-1658941800) snippet running without (obvious) error.

## Todos
  - [x] Include `VirtualSiteHandler` in system subset
  - [x] Include `VirtualSiteHandler`'s dependencies in system subset
  - [x] Add test(s)
    - [x] Halving (`scale_amount=-0.5`) `distance` parameter in OPC
    - [x] Nothing weird happens mixing molecules with and without virtual sites applied
    - [x] OpenMM energies can be evaluated

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/openff/evaluator/utils/openmm.py
+++ b/openff/evaluator/utils/openmm.py
@@ -199,7 +199,6 @@ def system_subset(
         # conformer's geometry likely does not match the force field geometry
         handlers_to_register.update(
             {
-                "vdW",
                 "Bonds",
                 "Constraints",
             },
--- a/openff/evaluator/utils/openmm.py
+++ b/openff/evaluator/utils/openmm.py
@@ -199,6 +199,7 @@ def system_subset(
         # conformer's geometry likely does not match the force field geometry
         handlers_to_register.update(
             {
+                "vdW",
                 "Bonds",
                 "Constraints",
             },
--- a/openff/evaluator/utils/openmm.py
+++ b/openff/evaluator/utils/openmm.py
@@ -179,21 +179,28 @@ def system_subset(
 
     handlers_to_register = {parameter_key.tag}
 
-    if parameter_key.tag in {"ChargeIncrementModel", "LibraryCharges"}:
-        # Make sure to retain all of the electrostatic handlers when dealing with
-        # charges as the applied charges will depend on which charges have been applied
-        # by previous handlers.
-        handlers_to_register.update(
-            {"Electrostatics", "ChargeIncrementModel", "LibraryCharges"}
-        )
-
-    if parameter_key.tag in {"VirtualSites"}:
+    if parameter_key.tag in {"ChargeIncrementModel", "LibraryCharges", "VirtualSites"}:
+        # Make sure to retain _all_ of the electrostatic handlers when dealing with any handler
+        # that might modify changes as the applied charges will depend on which charges have been
+        # applied by previous handlers.
         handlers_to_register.update(
             {
                 "Electrostatics",
-                "vdW",
-                "LibraryCharges",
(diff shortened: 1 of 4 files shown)


jthorton:
I think we could use something like this to include all standard electrostatic forces, tagging @SimonBoothroyd to double to check.
```python
if parameter_key.tag in {"ChargeIncrementModel", "LibraryCharges", "VirtualSiteHandler"}:
        # Make sure to retain all of the electrostatic handlers when dealing with
        # charges as the applied charges will depend on which charges have been applied
        # by previous handlers.
        handlers_to_register.update(
            {"Electrostatics", "ChargeIncrementModel", "LibraryCharges", "VirtualSiteHandler", "ToolkitAM1BCC"}
        )
```

SimonBoothroyd:
@jthorton that looks right - good catch on `"ToolkitAM1BCC"` as I think that should be there also!

mattwthompson:
For testing, would it be sufficient to make system subsets in a number of different cases or should we set up some actual gradient-based calculations?

openff/evaluator/utils/openmm.py
```diff
@@ -179,12 +179,30 @@ def system_subset(
...
+            }
+        )
+
+    if parameter_key.tag in {"VirtualSites"}:
+        # Interchange's current implementation uses bonds and constraints to determine the values
+        # OpenMM needs for virtual sites; using positions might not produce accurate results since a
+        # conformer's geometry likely does not match the force field geometry
+        handlers_to_register.update(
+            {
+                "vdW",
```
jthorton, line 202: I don't think we need the vdW handler if we take the gradient of an electrostatic parameter, I am not sure how much of a performance difference this would actually make however.

jthorton:
> For testing, would it be sufficient to make system subsets

I think this would be enough and maybe make sure we can calculate the energy of the sub system using openmm and matches what we expect particularly for systems with v-sites, maybe add in a system with v-site water and some other molecule without a v-site.

openff/evaluator/utils/openmm.py
jthorton: Ahhh no my bad I forgot we allow v-sites to have vdw interactions as well so I think you were correct sorry about that!

mattwthompson: I was a little confused - IIUC `vdWHandler` and `VirtualSiteHandler` completely separate the interactions, i.e. one handles all of the vdW interactions on the atoms and the other only on the virtual sites, no interaction or cross-talk like happens with charge increments, etc. in electrostatics. But then I guess you actually need the atomic vdW to evaluate the virtual sites' vdW since the system subset is slicing the system by interaction - in other words, you'd need all vdW interactions whether or not you're fitting all of them.

Is this coherent at all? I think we do need `"vdW"` included in this block, so if you also do, it's okay if I'm not making sense here.

jthorton: Exactly we need to include all handlers which might touch that interaction we are fitting to and as our v-sites can hit both vdW and electrostatics we need to be careful to include them. So I agree they should be in this block. 

We don't actually use v-sites with vdW parameters yet but if we do I can see the reverse also being an issue. Where we fit vdW and include v-sites with charges and LJ terms, we would only have charges on the atoms with v-sites and nothing else.

jthorton (commented):
LGTM!

jthorton (approved):

mattwthompson:
Thanks!

PR #535, opened 2023-10-13
Fix typo from #533

mattwthompson:
@aehogan I think this should be sufficient? Let me know if there's anything else I've goofed


aehogan (approved):
Looks good

PR #496, opened 2023-02-09
Remove protocols based on Yank

mattwthompson:
## Description
#488

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [ ] Remove protocols based on Yank
  - [ ] Remove dependencies only used by Yank protocols
  - [ ] Update release history

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

PR #543, opened 2023-11-13
Add back FreeSolv curation removed in #496

mattwthompson:
## Description
In #496, I removed the FreeSolv curation module, I think a mistake. There's no replacement for Yank, but that doesn't necessarily mean the data provider should be removed.

## Todos
  - [ ] Add back FreeSolv curation

## Status
- [ ] Ready to go


mattwthompson:
Okay, this is not so straightforward. The class needs `SolvationFreeEnergy` to exist - maybe it can be added back without properly being able to run.

PR #569, opened 2024-09-04
Scale MW back by Avogadro's number

### LINKED ISSUE #568: _PureOrMixtureData._smiles_to_molecular_weight gives incorrect MW
**Describe the bug**

Unfortunately this method does not return the MW for a single molecule but instead divides the MW by `unit.mole`, resulting in a per-mol weight that's scaled by Avogadro's number.

https://github.com/openforcefield/openff-evaluator/blob/c8c476ab6ec1e379049384ce23982e547d699f4b/openff/evaluator/datasets/thermoml/thermoml.py#L857-L859

This is the root cause of #553 -- while `_convert_mass_fractions` had units changed presumably to accomodate this, `_convert_molality` did not.

lilyminium:
## Description
Fixes #568 

## Todos
  - This scales the MW back and adds more granular tests.
  - It also adds some `__repr__` functions I used while debugging, which can be removed but might be nice for future debugging!

### DIFF
--- a/openff/evaluator/_tests/test_datasets/test_thermoml.py
+++ b/openff/evaluator/_tests/test_datasets/test_thermoml.py
@@ -5,12 +5,18 @@ Units tests for openff.evaluator.datasets
 import pytest
 from openff.units import unit
 
+import numpy as np
+
+
 from openff.evaluator.attributes import UNDEFINED
 from openff.evaluator.datasets import PhysicalProperty, PropertyPhase
 from openff.evaluator.datasets.thermoml import thermoml_property
 from openff.evaluator.datasets.thermoml.thermoml import (
     ThermoMLDataSet,
     _unit_from_thermoml_string,
+    _PureOrMixtureData,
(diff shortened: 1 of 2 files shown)


lilyminium:
Ah, forgot about forks and OpenEye... 

On the runs without OpenEye and without the fix, [the tests of converting solvent mole fractions and molality fail](https://github.com/openforcefield/openff-evaluator/actions/runs/10701761303/job/29668417768?pr=569), as well as the slightly stronger test of loading a ThermoMLDataset from an XML file with molality constraints (`test_thermoml_molality_constraints`) and checking multiple components are parsed.

lilyminium:
Apologies for the failing OpenEye tests, but I think this is ready for review!

mattwthompson (approved):
LGTM, I'm kicking this patch upstream to be safe

mattwthompson:
Merged #570

lilyminium:
Thanks Matt!

PR #577, opened 2024-10-01
Fix mole fractions

lilyminium:
## Description

PR #576 but from main repo so tests can run.

I've added tests for the default_simulation_schema, but the default_reweighting_schema also uses the same incorrect substance.  I don't have adequate data and understanding of the structure of the reweighting protocol to add similar tests for it, and given how long this PR took, I estimate it would take a couple days. My proposal is to implement the same fix as was tested in `default_reweighting_schema` and raise an issue to generate more tests later, but I'll defer to the infrastructure team here.

## Done
- [x] Added tests
- [x] added fix

## Todos
  - [ ] Reweighting tests
  - [x] Reweighting fix


lilyminium:
If of any interest: one point of friction with the pre-commit pushing into my PR that I'm noticing is that it forces me to continuously `git stash` my current work, pull and rebase, and reapply, as I frequently only push part of the codebase at a time to a PR (e.g. in the example of wanting to ensure tests fail across a CI matrix before committing and pushing the fix for them). Maybe this is a very niche way of doing things, though.

lilyminium:
Thanks for the tip @mattwthompson!

Pending tests finishing, I think this is ready for review. [These tests](https://github.com/openforcefield/openff-evaluator/actions/runs/11125053754/job/30936405728?pr=577) show the expected failure when the test isn't put in, i.e. the mole fractions are incorrect. There are two tests that are run: one that directly only checks mole fractions, and one that uses @barmoral's data to set up a more detailed test that executes most steps in the protocol except the ones requiring simulation. It also explicitly checks that the mole fraction weighted values are as expected.

openff/evaluator/_tests/test_properties/test_enthalpy.py
```diff
@@ -0,0 +1,354 @@
...
+        constructed to test expected mole fractions. See Issue #575.
+
+        This test steps through the EnthalpyOfMixing process.
+        Some data is provided to avoid re-simulating long trajectories:
+
+        - energy minimisations and equilibrations
+        - the production simulation observables
+        - the decorrelated trajectory and observables
+
+        Note: the test is expected to take a few minutes to run.
```
mattwthompson, line 206: We could move these to GPUs at some point if this is too slow for CI

mattwthompson (approved):

mattwthompson:
Thanks for patching this - I think we have plenty of ideas for more work we could do, but this fixes the more discrete issue that Barbara reported, so I'm just going to merge this and make a release now.

openff/evaluator/_tests/test_properties/test_enthalpy.py
lilyminium: Sounds good -- it is doable right now but definitely one of the slower tests in the codebase.

Issue #587, opened 2024-12-03
Use LangevinMiddleIntegrator for OpenMM simulations


lilyminium:
**Is your feature request related to a problem? Please describe.**

Evaluator currently uses `openmmtools.integrators.LangevinIntegrator`. IMO we should probably be using the [LangevinMiddleIntegrator](http://docs.openmm.org/latest/api-python/generated/openmm.openmm.LangevinMiddleIntegrator.html) which provides more accurate sampling. I also think I might have traced an odd issue where the integrator wasn't correctly regulating the temperature to 298K to the openmmtools one, which produces an average temperature of <290K; the openmm version correctly yields are temperature that fluctuates around ~300 K. (see attached screenshots).

Screenshots of properties correspond to:
* equilibration_statistics_openmmtools: `openmm.integrators.LangevinIntegrator`, v0.23.1
* equilibration_statistics_openmm: `openmm.openmm.LangevinIntegrator`, 8.1 version
* equilibration_statistics_middle: `openmm.openmm.LangevinMiddleIntegrator`, 8.2 version (in 8.2, LangevinIntegrator and LangevinMiddleIntegrator are the same)

**OpenMMTools, LangevinIntegrator**
![equilibration_statistics_openmmtools](https://github.com/user-attachments/assets/b69e5b5c-db9c-433c-8d72-aea108b33e37)

**OpenMM, LangevinIntegrator**
![equilibration_statistics_openmm](https://github.com/user-attachments/assets/158dec8f-c6c4-4bc4-86ae-9664a8124e92)
github.com/user-attachments/assets/cb614ab2-5cbf-467c-aba8-c94264a29141)

**OpenMM, LangevinMiddleIntegrator**
![equilibration_statistics_middle](https://github.com/user-attachments/assets/cf0d075c-bfb5-4849-a701-c1bbbea55b03)

mrshirts:
Ouch! Yes, definitely should switch. The new KE looks right. You should check the distribution of the kinetic energy matches the correct Maxwell-Boltzmann distribution to within reasonable noise - https://github.com/shirtsgroup/physical_validation provides some quick tools to do this.

mattwthompson:
Do the input arguments map on 1:1? I see something in inverse time that defaults to 1 inverse picosecond in each class, but they're named differently:

[...]

PR #588, opened 2024-12-03
Use `openmm.LangevinMiddleIntegrator`

mattwthompson:
## Description
Implements #587

## Todos
  - [x] `openmmtools.integrators.LangevinIntegrator` -> `openmm.LangevinMiddleIntegrator`

## Questions
- [ ] Arguments actually a drop-in replacement?

## Status
- [ ] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -8,6 +8,11 @@ Releases follow the ``major.minor.micro`` scheme recommended by
 * ``minor`` increments add features but do not break API compatibility
 * ``micro`` increments represent bugfix releases or improvements in documentation
 
+Current development
+-------------------
+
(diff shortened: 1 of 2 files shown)


lilyminium:
> Arguments actually a drop-in replacement?

I *think* they should be -- not only are the units and default argument the same, but the collision_rate is converted to `self._gamma` in openmmtools, where gamma is typically used to represent the friction coefficient.

lilyminium (approved):
LGTM -- thanks for the quick fix @mattwthompson!

My only thought -- does this need tests of some kind, e.g. to check the temperature issue? But that might be out of scope and a lot more work for a one-line change, happy to defer to you.

mattwthompson:
Agree it would be useful but I don't know a quick way to test that right now

Issue #593, opened 2025-01-06
Add an equilibration phase for slow-equilibrating properties


lilyminium:
Related to #591. As discussed in the FF fitting meeting and on Slack, one solution to slow-equilibrating properties would be have an "iteration 0" in FF optimization and benchmarking that allows the use of previously simulated boxes to provide initial coordinates.

This solution should ideally:
- generalize broadly and robustly to unseen properties (i.e. there should be no danger of similar but different properties contaminating a run)
- generalize across force fields
- allow use of previously-computed data outside of the current Evaluator run
- work across optimization and benchmarking

lilyminium:
Going to close this as #607 has been merged and all subissues closed.

Issue #592, opened 2025-01-05
Relative uncertainty does not result in successful properties unless relative_uncertainty > 1.0


lilyminium:
**Describe the bug**

When determining whether a property has been successfully computed using the SimulationLayer, Evaluator checks whether the uncertainty/error of the property is within either the user-specified absolute_tolerance or relative_tolerance. However, as written, the relative_tolerance will only successfully terminate if it's >1. I think this may be due to specifying the tolerance to check incorrectly.

**To Reproduce**

```
#!/usr/bin/env python
# coding: utf-8

# In[4]:


from openff.units import unit
from openff.evaluator.backends import ComputeResources
from openff.evaluator.backends.dask import DaskLocalCluster
from openff.evaluator.datasets import (
    MeasurementSource,
    PhysicalPropertyDataSet,
    PropertyPhase,
)
from openff.evaluator.client import EvaluatorClient, RequestOptions, Request, ConnectionOptions
from openff.evaluator.server.server import Batch, EvaluatorServer

from openff.evaluator.forcefield import (
    LigParGenForceFieldSource,
    SmirnoffForceFieldSource,
    TLeapForceFieldSource,
)

from openff.evaluator.properties import Density, EnthalpyOfMixing
from openff.evaluator.substances import Substance
from openff.evaluator.thermodynamics import ThermodynamicState


dataset = PhysicalPropertyDataSet()
thermodynamic_state = ThermodynamicState(
    temperature=298.15 * unit.kelvin,
    pressure=101.325 * unit.kilopascal,
)
dataset.add_properties(
    Density(
        thermodynamic_state=thermodynamic_state,
        phase=PropertyPhase.Liquid,
        value=1.0 * Density.default_unit(),
        uncertainty=1.0 * Density.default_unit(),
        source=MeasurementSource(doi=" "),
        substance=Substance.from_components("CO"),
    ),
)
for i, prop in enumerate(dataset.properties, 1):
    prop.id = str(i)


options = RequestOptions()
options.calculation_layers = ["SimulationLayer"]
schema = Density.default_simulation_schema(
    n_molecules=256,
    relative_tolerance=0.5
)

options.add_schema(
    "SimulationLayer",
    "Density",
    schema,
)


force_field_path = "openff-2.1.0.offxml"
force_field_source = SmirnoffForceFieldSource.from_path(
    force_field_path
)



with DaskLocalCluster(
    # uncomment options below to use a GPU to compute (much faster than CPU-only).
    
[...]

PR #594, opened 2025-01-06
Add Equilibration and PreequilibratedSimulation phases

lilyminium:
## Description
Addresses #593 

This PR adds EquilibrationLayer and PreequilibratedSimulationLayers to Evaluator as one solution to #593. The envisioned use is:

- running properties through an `EquilibrationLayer` to equilibrate within a particular potential energy tolerance
- Use the equilibrated boxes as input to a PreequilibratedSimulationLayer, which runs through all steps of a regular SimulationLayer *except* box-packing.

This PR branches off #589. 

Not currently addressed (perhaps in another PR?):
- relative_tolerance is unsupported for `EquilibrationLayer` due to its application as `relative_tolerance * measured_property.uncertainty`, which requires the units to match the original property. Instead, the `absolute_tolerance` can be specified in kJ/mol for potential energy.
- relative_tolerance may not result in successful properties (issue #592)
- properties that aren't Densities or EstimatibleExcessProperties (and only EnthalpyOfMixing has actually been tested) are still unsupported, although it should be easy to extend
- the equilibrated boxes must be present in the storage backend at the moment instead of being in a separate directory. They're not very human-readable either.
- PreequilibratedSimulationLayer requires pre-equilibrated boxes to have been computed with a `EquilibrationLayer` beforehand (the storage query should check for this in a hardcoded way), to avoid contamination from other stored data.
- have not yet tested re-use of equilibrated boxes across range of force fields
- have not yet figured out what should happen if multiple boxes are present per system in the stored data. Right now there's a hard-coded check for only finding one box. This should get replaced with something that sorts results somehow and picks one.
- It would be ideal to have the EquilibrationLayer also query the storage backend and exit early if a result already exists.

TODO:
- [ ] a full test run with properties of concern, and benchmark GPU performance on k8s.
- [ ] update the tutorial notebook once it has finished executing slowly on my laptop


lilyminium:
Closing for now as this is still very early stages.

mattwthompson:
This was looking good, but I'll let you decide where and when you want this to go

PR #602, opened 2025-01-20
Add NoBatch mode

### LINKED ISSUE #601: Add NoBatch mode
The EvaluatorServer batches properties for computing and assigns each batch a unique UUID, which is nice for computing things in smaller chunks. It's less nice for writing and running tests. Would be great to have a NoBatch mode that puts everything in a single batch with reproducible naming for testing reasons.

lilyminium:
## Description
Fixes #601 

- Adds a mode that just creates batches that increment sequentially
- Refactors tests slightly to use fixtures and take place in a temporary directory -- running an EvaluatorServer creates a persistent `stored_data` directory

## Status
- [ ] Ready to go

### DIFF
--- a/openff/evaluator/server/server.py
+++ b/openff/evaluator/server/server.py
@@ -266,6 +266,27 @@ class EvaluatorServer:
             request_results.exceptions.extend(batch.exceptions)
 
         return request_results, None
+    
+    def _no_batch(self, submission, force_field_id):
+        """Returns a single Batch."""
+
+        reserved_batch_ids = {
+            *self._queued_batches.keys(),
+            *self._finished_batches.keys(),
+        }
+        batch = Batch()
+        batch.force_field_id = force_field_id
+        batch.enable_data_caching = self._enable_data_caching
+        batch.queued_properties = list(submission.dataset.properties)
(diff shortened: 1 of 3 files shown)


openff/evaluator/server/server.py
```diff
@@ -434,6 +457,8 @@ def _prepare_batches(self, submission, request_id):
             batches = self._batch_by_same_component(submission, force_field_id)
         elif batch_mode == BatchMode.SharedComponents:
             batches = self._batch_by_shared_component(submission, force_field_id)
+        elif batch_mode == BatchMode.NoBatch:
+            batches = self._no_batch(submission, force_field_id)
```
mattwthompson, line 461: Could you add a test that covers these lines?

mattwthompson (changes requested):

openff/evaluator/server/server.py
lilyminium: Sure -- added a test for _prepare_batches.

lilyminium: PS This code path should be used in the tests in #604 and #608.

mattwthompson (approved):
LGTM

PR #604, opened 2025-01-21
Fix relative tolerance

### LINKED ISSUE #592: Relative uncertainty does not result in successful properties unless relative_uncertainty > 1.0

lilyminium:
## Description
Fixes #592

Changes:
* changes which uncertainty is compared
* Adds tests of directly executing a Simulation graph, as well as with a server
* Adds test data from an example run
* Adds some helper functions to tests

Blocked by / follows #602.

## Questions
- I've started adding test data from example runs -- this is helpful firstly in having an example of test data, and secondly for avoiding re-execution of workflows during tests. However there are a lot of files -- should these be compressed and uncompressed on-the-fly? Pros for compression are that the number of files changed is much lower. Cons is that the tarball may wind up being quite large (10+MB), and any changes to an individual file will change the entire compressed directory.

## Status
- [x] Ready to go


mattwthompson (approved):
Mild preference for compressing data files since they're already so large, but I think we should go with whatever option slows you down the least

lilyminium:
Thanks Matt! I'll merge this in for now. I'm about to add a whole lot more test data to some of my other PRs so we might want to revisit better test data storage solutions there, but let's see how bad it is first :-)

PR #606, opened 2025-01-21
Add AttributeClass initialization

### LINKED ISSUE #605: Allow AttributeClasses to be initialized with values
Currently all subclasses of AttributeClasses need their attributes to be set separately to instantiation, e.g.:

```
batch = Batch()
batch.id = ...
batch.force_field_id = ...
```

It'd be quite handy to be able to initialize attributes e.g. `Batch(id=..., force_field_id=...)`.

Some classes define `__init__` methods to allow this, so there may be a reason why this isn't more generally supported?

lilyminium:
Fixes #605

### DIFF
--- a/openff/evaluator/_tests/test_attributes/test_attributes.py
+++ b/openff/evaluator/_tests/test_attributes/test_attributes.py
@@ -114,3 +114,20 @@ def test_nested_validation():
 
     nested_object.some_dict["x"].required_input = ""
     nested_object.validate()
+
+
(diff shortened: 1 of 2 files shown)


mattwthompson (approved):
Fine by me, I trust this enables some of your work that you'd otherwise find clunky in a way you don't prefer. And this is a backwards-compatible change, i.e. the "instantiate and set" pattern is unaffected

lilyminium:
Thanks Matt!

PR #611, opened 2025-01-22
Add easy way to create new substances for real numbers of mols

### LINKED ISSUE #610: Add easy way to create new substance for specific number of molecules
A Substance may be created with mole fractions of 0.5098 for one component and 0.4902 for another, but when actually creating a box of 100 molecules, this gets turned into actual mole fractions of 0.51 and 0.49 respectively. This will pose an issue when querying storage backends for existing equilibrated boxes, as what's in the storage backend (i.e. the latter, simpler mole fractions) won't match the exact substance in the query. Instead, there should be an easy way to create a new, simpler substance matching the total number of molecules for the query.

lilyminium:
## Description
Fixes #610 

Largely just copies very similar code from the BuildCoordinatesPackmol protocol; possibly these could be merged in some way.

### DIFF
--- a/openff/evaluator/_tests/test_substances.py
+++ b/openff/evaluator/_tests/test_substances.py
@@ -98,3 +98,14 @@ def test_truncate_n_molecules():
 def test_substance_len():
     substance = Substance.from_components("C", "CC", "CCC", "CCC")
     assert len(substance) == 3
+
+
+def test_to_substance_n_molecules():
+    substance = Substance()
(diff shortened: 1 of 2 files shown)


mattwthompson (approved):
LGTM

I looked at the method you refer to and agree that some refactoring would be useful, but that doesn't need to be done right now

lilyminium:
Thanks Matt!

PR #626, opened 2025-03-04
Fail individually on bad SMILES

### LINKED ISSUE #620: Evaluator should fail individually on bad SMILES from ThermoML instead of entire file
**Describe the bug**

When Evaluator parses a ThermoML file, it tries to read SMILES. If the SMILES fails (e.g. on a radical error), the exception means the rest of the file is also ignored, instead of that singular entry. Due to differences in the OpenFF toolkit over different versions (e.g. the addition of the radical error), this means some past work is unnecessarily highly unreproducible (e.g. the filtering of the Sage training dataset for 2.0). Beyond reproducibility concerns, this also impacts the reading of any properties later in files that contain invalid molecules, which I ran into while trying to read viscosities.

**Desired solution**

Wrap the call below, or the individual smiles call within it, in a try/except:

https://github.com/openforcefield/openff-evaluator/blob/7d927d9a0a527c9833c2124c0db85af5a1c8e999/openff/evaluator/datasets/thermoml/thermoml.py#L2187-L2188

**Output**

 - Operating system

**Additional context**

FWIW I don't think (~70% sure) that fixing this issue will make the Sage filtering 100% reproducible, but... less unreproducible.

lilyminium:
## Description
Fixes #620

Changes:
* adds a bad SMILES to an existing data file to make a test fail
* Wraps Compound creation in a try/except

### DIFF
--- a/openff/evaluator/_tests/test_datasets/test_thermoml.py
+++ b/openff/evaluator/_tests/test_datasets/test_thermoml.py
@@ -193,12 +193,19 @@ def test_thermoml_molality_constraints():
     assert len(data_set.properties[0].substance) > 1
 
 
-def test_thermoml_mole_constraints():
+def test_thermoml_mole_constraints(caplog):
     """A collection of tests to ensure that the Mole fraction constraint is
     implemented correctly alongside solvent constraints."""
 
     # Mole fraction
-    data_set = ThermoMLDataSet.from_file(get_data_filename("test/properties/mole.xml"))
+    # This file contains a bad smiles to test Issue #620
+    # Test that the file a) gets parsed
+    # and b) logs a warning about parsing radicals
+
+    with caplog.at_level("WARNING"):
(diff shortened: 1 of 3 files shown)


mattwthompson (changes requested):
Perfect, I knew it'd be a small LOC fix but I wasn't sure where

---

The info somebody else is all here in the history of the commit and tagged issue, but to better help future people remember this change, could you

1. Add a note of this behavior into the docstring
1. Make sure that the logger warning is captured in the test (I think [here](https://github.com/openforcefield/openff-evaluator/blob/b507e0581396e06389446e39d925fc54a4d345f1/openff/evaluator/_tests/test_datasets/test_thermoml.py#L201) is the only time this file is touched?)

lilyminium:
Haha, this is very much the easy option I mentioned in the issue as I didn't have the time to try a more fine-tuned approach. I've updated the docstring and test!

mattwthompson (approved):

PR #638, opened 2025-03-26
Add PreequilibratedSimulationLayer

lilyminium:
## Description
Follows on from #608 to add a PreequilibratedSimulationLayer for simulating from existing workflows.


openff/evaluator/layers/equilibration.py
```diff
@@ -82,6 +82,16 @@ class EquilibrationProperty(AttributeClass):
...
 
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return False
+
+        cls_attributes = self.get_attributes()
+        for attribute_name in cls_attributes:
+            if getattr(self, attribute_name) != getattr(other, attribute_name):
+                return False
+        return True
```
lilyminium, line 93: Can this be upstreamed to AttributeClass?

It would be nice to compare objects easily, but I'm not sure how many checks there are in the code for identity vs equality.

openff/evaluator/protocols/utils.py
```diff
@@ -377,71 +396,19 @@ def generate_reweighting_protocols(
     return protocols, data_replicator
 
 
-def generate_equilibration_protocols(
+def generate_conditional_equilibration_protocols(
```
lilyminium, line 399: This could have been diffed smarter. This essentially moves a solid chunk of code out of `generate_equilibration_protocols` so both `generate_equilibration_protocols` and `generate_preequilibrated_simulation_protocols` can generate conditional equilibration groups.

openff/evaluator/properties/density.py
```diff
@@ -105,6 +109,76 @@ def default_equilibration_schema(
         calculation_schema.workflow_schema = schema
         return calculation_schema
 
+    @classmethod
+    def default_preequilibrated_simulation_schema(
```
mattwthompson, line 113: IIUC this or something similar should also be copied into `openff/evaluator/properties/dielectric.py`?

openff/evaluator/properties/properties.py
```diff
@@ -192,6 +196,224 @@ def default_equilibration_schema(
         )
         return calculation_schema
 
+    @classmethod
+    def default_preequilibrated_simulation_schema(
```
mattwthompson, line 200: I tried to reason through how some of this partially-duplicated code could be consolidated down, but didn't get anywhere

openff/evaluator/layers/equilibration.py
mattwthompson: Probably, but I'm inclined to suggest that we do that in a follow-up

mattwthompson (approved):
I think our strategy here ought to be similar to the last big PR; merge more or less in the current state, "test" it again by running some fits, and iterate from there. I looked through the source code and found nothing I didn't like and only one question, so I'm "approving" this modulo whatever changes you wish to make to it. I would still like you to do a once-over through the data files that were added to make sure nothing there was added inadvertently (I see a SLURM logs and a notebook - if that's handy or used for provenance, no issue at all). But if/once you're happy with it and it's working in the ways you expect for the current scope, please hit the big green button

openff/evaluator/properties/density.py
lilyminium: Probably, but I haven't been computing and testing dielectrics (or any non-enthalpies of mixing + densities) so was thinking of adding this in a future version. Would it be better to add a method that raises a `NotImplementedError` here?

openff/evaluator/properties/properties.py
lilyminium: I think just passing in the input schema object and generating protocols from there would be an improvement.

There are other things that'd be great to fix, like how hard it is right now to change details like the simulation runtime from the schema that's output from this method... but things that can also happen in the future.

lilyminium:
Thanks for the review @mattwthompson! I've removed the slurm run file, but I might be blind because I can't seem to find the notebook at all -- do you remember where you saw that? There is a Python file that's formatted like a notebook that was used to generate test data, which I'd suggest keeping as it would come in handy in the future for regenerating data if any attributes are added/modified.

mattwthompson:
Whoops, I meant this file:

> openff/evaluator/data/test/workflows/preequilibrated_simulation/dhmix-density-CCCO/test-preeq.py

since it looks like it was previously a notebook and converted to a script, and located in an odd location

lilyminium:
Ah yep that's the one I meant, it's useful as a way to re-generate data (and it's in that directory to generate the data in the directory) so I'd recommend keeping it, although happy to move it elsewhere if that location is too odd.

mattwthompson:
Good to me as-is

Issue #660, opened 2025-05-15
Heat of vap training targets failing


JHoeflich1:
I'm trying to optimize LJ parameters for a set of alkanes using a custom training dataset containing heat of vaporization and density targets. I followed tutorial01 to create my dataset and tutorial04 to set up the ForceBalance optimization inputs.

The job runs on my university SLURM cluster, but a number of "conditional groups" fail with this error:

File "/projects/juho8819/software/anaconda/envs/test_evaluator_oe/lib/python3.11/site-packages/openff/evaluator/workflow/protocols.py", line 1201, in _execute_protocol
    protocol.execute(directory, available_resources)
  File "/projects/juho8819/software/anaconda/envs/test_evaluator_oe/lib/python3.11/site-packages/openff/evaluator/workflow/protocols.py", line 688, in execute
    self._execute(directory, available_resources)
  File "/projects/juho8819/software/anaconda/envs/test_evaluator_oe/lib/python3.11/site-packages/openff/evaluator/protocols/openmm.py", line 549, in _execute
    _compute_gradients(
  File "/projects/juho8819/software/anaconda/envs/test_evaluator_oe/lib/python3.11/site-packages/openff/evaluator/protocols/openmm.py", line 212, in _compute_gradients
    reverse_energies = _evaluate_energies(
                       ^^^^^^^^^^^^^^^^^^^
  File "/projects/juho8819/software/anaconda/envs/test_evaluator_oe/lib/python3.11/site-packages/openff/evaluator/protocols/openmm.py", line 112, in _evaluate_energies
    update_context_with_positions(openmm_context, positions, box_vectors)
  File "/projects/juho8819/software/anaconda/envs/test_evaluator_oe/lib/python3.11/site-packages/openff/evaluator/utils/openmm.py", line 326, in update_context_with_positions
    context.setPeriodicBoxVectors(*box_vectors)
  File "/projects/juho8819/software/anaconda/envs/test_evaluator_oe/lib/python3.11/site-packages/openmm/openmm.py", line 11292, in setPeriodicBoxVectors
    return _openmm.Context_setPeriodicBoxVectors(self, a, b, c)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #669, opened 2025-06-02
Fudge box vectors with MDTraj/OpenMM DCD issue

mattwthompson:
## Description
Plausible fix for #660 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -8,11 +8,15 @@ Releases follow the ``major.minor.micro`` scheme recommended by
 * ``minor`` increments add features but do not break API compatibility
 * ``micro`` increments represent bugfix releases or improvements in documentation
 
+Current development
+-------------------
+
+* PR `#669 <https://github.com/openforcefield/openff-evaluator/pull/669>`_: Work around a wart in MDTraj in which orthogonal boxes are sometimes reported as triclinic.
+
 0.5.0 - April 30, 2025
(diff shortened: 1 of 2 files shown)


Yoshanuikabundi (approved):
I think this is my preferred option. I'd be inclined to interpret this issue as "There's a bug in mdtraj, the tolerance for the box fudging isn't large enough", for which the fix is to "patch" calls to that code with our own fudging with a larger tolerance. This avoids adding any extra logic, and I suspect will make similar errors less likely in the future. I'm not sure what similar errors would look like because this failure mode also relies on a somewhat esoteric interaction with how OpenMM implements pressure coupling, but this fix is conservative enough that I think it's appropriate.

I have two suggestions:

1. Apply this fix to other uses of `trajectory.openmm_boxes(...)` - my grepping just found one other use, in line 190/216 (before/after this change) of this same file
2. Very optionally - refactor `_fudge_openmm_box(trajectory.openmm_boxes(...` into something like `openmm_box_from_trajectory(trajectory, frame_index)` - This should be a very minor change but I think it encapsulates the logic a bit better, and it means that this "patch" for an "mdtraj bug" is co-located with all calls to the "buggy" code.

mattwthompson:
Thanks @Yoshanuikabundi!

Issue #661, opened 2025-05-22
Support `pymbar` 4


mattwthompson:

mattwthompson:
It's not clear how much value this is since ForceBalance is not being updated for compatibility

PR #678, opened 2025-07-14
Test on Python 3.13

### LINKED ISSUE #679: Support Python 3.13

mattwthompson:
Trying to support Python 3.13, would fix #679


mattwthompson:
Blocked by https://github.com/openforcefield/openff-evaluator/issues/661 and Foyer (which should be shuffled to an optional dependency)

PR #682, opened 2025-07-28
Prototype hash function

### LINKED ISSUE #658: Unique meaningful hash ID for PhysicalProperties
**Is your feature request related to a problem? Please describe.**

Can PhysicalProperties be assigned a unique ID based on the hash of their contents? Currently I believe when downloaded from ThermoML they get a random UUID which is difficult to compare between datasets.

**Describe the solution you'd like**

Hash a tuple of (property type, substance, thermodynamic state, value, error, source), and any other attrs/columns that distinguish two properties. This will run into false mismatches from precision differences but it will be much more comparable than random UUIDs.

**Describe alternatives you've considered**

**Additional context**

lilyminium:
## Description
Fixes #658 

This hash function includes: substance, phase, thermodynamic state, value, uncertainty, source
It ignores:
- `id`: as mentioned in #658, the ID attr is currently mostly meaningless (a unique ID per property that unfortunately changes between runs)
- `metadata`: I think this is used during property computation? From a quick grep I'm not sure this is actually used outside of taproom.py
- `gradients`: used in FF fitting

To do: tests, but would like to canvas opinions before moving forward

### DIFF
--- a/openff/evaluator/datasets/datasets.py
+++ b/openff/evaluator/datasets/datasets.py
@@ -228,7 +228,7 @@ class PhysicalProperty(AttributeClass, abc.ABC):
         }
         serialized = json.dumps(obj, sort_keys=True, cls=TypedJSONEncoder)
         return int(hashlib.sha256(serialized.encode("utf-8")).hexdigest(), 16)
-    
+
     def get_property_hash(self) -> int:
         """
         Returns a hash of the property based on attributes that are expected to


lilyminium:
I forgot that Python 3.3+ randomizes hashes, which complicates things more if we want hashes to be comparable between runs, since a number of functions currently use the inbuilt `hash()`. We would need some kind of solution that uses `hashlib` hashing of a serialized representation for consistent hashing between Python runs, I think.

jameseastwood:
@mattwthompson can you take a look at Lily's PR and see if you notice any obvious dangers or concerns before she continues with the implementation?

mattwthompson:
No objections here

Having not run many fits, most of my context comes from reading through the linked issue. If property-based hashes are more useful than ThermoML's UUIDs, then I see the motivation for the change. There are probably some details to be fiddled with re: which attributes to include in the hash, but that can be figured out in the process of implementing and using this change.

PR #685, opened 2025-08-04
Fix filtering schema validators

### LINKED ISSUE #687: Validators of some filtering schemas broken

mattwthompson:
## Description
Closes #687

## Todos
  - [x] Fix some validators in filtering
  - [x] Add a minimal test

## Status
- [ ] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -8,6 +8,11 @@ Releases follow the ``major.minor.micro`` scheme recommended by
 * ``minor`` increments add features but do not break API compatibility
 * ``micro`` increments represent bugfix releases or improvements in documentation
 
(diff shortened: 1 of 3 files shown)


lilyminium (approved):
LGTM, although would suggest changing `cls` to `self` to avoid confusion. Thanks for looking into this Matt!

I realise #686 is still open but it'd be great if the test did ultimately include that filter, as existing FF development code for Sage 2.0 uses a full workflow of filtering and selection filters.

mattwthompson:
Renaming actually breaks it, for reasons I'm not able to wrap my head around. Just tagging them as class methods again gets things passing, but that might not be the best solution.

PR #732, opened 2026-02-11
Update ParameterGradientKeys for vsite key matching

### LINKED ISSUE #726: Can't re-fit virtual sites due to inability to identify them uniquely
**Describe the bug**

Theoretically I should be able to re-fit virtual sites with Evaluator, but I can't because I can't retrieve a virtual site parameter by SMIRKS. This is by design as [SMIRKS does not uniquely identify a VirtualSite parameter](https://github.com/openforcefield/openff-toolkit/blob/1017880c2c63b4fbff5aa43b079508d50b987dd0/openff/toolkit/typing/engines/smirnoff/parameters.py#L1578-L1581) -- multiple are allowed to have the same SMIRKS. Therefore another solution is necessary. Interchange has VirtualSiteKeys.

**To Reproduce**

Apologies, I have no small MRE right now; I'm running into bugs while trialling a 4-site water model fit with ForceBalance. The problematic code is in system_subset:

https://github.com/openforcefield/openff-evaluator/blob/884b8504dbb9d8922909b0d2f2629234dc8bc7fc/openff/evaluator/utils/openmm.py#L238-L242

When the parameter key being fit is a virtual site, this triggers the NotImplementedError linked in the openff toolkit.

**Additional context**

While pushing through getting a 4-site fit off the ground, I quickly patched this code as shown here: https://github.com/openforcefield/openff-evaluator/commit/aa77c20a43c3bb1b1d2984ad8e880b651b4a0913

This is hacky as `ParameterKey.smirks` was probably not intended to hold all these attributes, but there were no other fields in ParameterKey to easily slot this on. It's also hardcoded to match corresponding patches I put in ForceBalance [here](https://github.com/leeping/forcebalance/commit/d4bb70bee660539ee26152776c92b21bba539c40) and [here](https://www.google.com/url?q=https://github.com/leeping/forcebalance/commit/bc2c3e101ac4c64b5cc7f818ad8e14e5b3b3cc16&sa=D&source=editors&ust=1770374696640310&usg=AOvVaw14imcz7LUcYmInyzIpE_-M).

This hardecoded vsite key was constructed to broadly approximate the Interchange version. The best solution would actually somehow incorporate a more canonical version of Interchange's VirtualSiteKey. However, I'm not certain the best way to, especially if `ParameterKey.smirks` needs to be a str.

lilyminium:
## Description
- Fixes #726
- updates ParameterGradientKeys to expect all required info for matching vsite keys
- Requires matching ForceBalance update

**note -- this may not be the best way to do it, i'm not sure if we can integrate canonical Interchange PotentialKeys instead**?

### DIFF
--- a/.github/workflows/aws.yaml
+++ b/.github/workflows/aws.yaml
@@ -35,10 +35,17 @@ jobs:
           GH_PAT: ${{ secrets.GH_PAT }}
   self-hosted-test:
     runs-on: self-hosted
+    timeout-minutes: 600 # 10 hours * 60 mins, also see the auto-shutdown step below for failsafe 
     needs:
       - start-aws-runner
     steps:
       - uses: actions/checkout@v6
+
+      - name: Schedule auto-shutdown after 24 hours 
+        # This is the failsafe if timeout-minutes above fails
+        # More info https://github.com/openforcefield/openff-evaluator/issues/740
+        run: nohup bash -c 'sleep 86400 && sudo shutdown -h now' > /tmp/auto-shutdown.log 2>&1 &                     
+
 
       - name: Print disk usage
         run: "df -h"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
   - id: black-jupyter
     files: ^docs/tutorials
 - repo: https://github.com/PyCQA/isort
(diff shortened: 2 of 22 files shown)


lilyminium:
Thanks Matt, yes #742 superseded it -- I'll close this PR.

PR #733, opened 2026-02-11
allow valence and nagl parameters

### LINKED ISSUE #727: Can't re-fit valence parameters in Evaluator, or vsites + NAGL
**Describe the bug**

Evaluator should be able to co-optimize valence parameters. Also, it should be able to re-fit virtual site charges where the base charge method is AshGC. The reason we can't do both of these things is broadly the same: Evaluator cannot calculate the energies of a system subset because it creates systems with no particles. (I think this is also the core issue plaguing the lipid cooptimizations).

**To Reproduce**

Apologies, I have no small MRE for this or for #726 yet. Both issues involve system_subset, so I will try to come up with something next week. The problematic code is in system_subset:

https://github.com/openforcefield/openff-evaluator/blob/884b8504dbb9d8922909b0d2f2629234dc8bc7fc/openff/evaluator/utils/openmm.py#L198-L224

**Output**

```
  File "/data1/choderaj/wangl14/pydev/openff-evaluator/openff/evaluator/workflow/protocols.py", line 688, in execute
  File "/data1/choderaj/wangl14/pydev/openff-evaluator/openff/evaluator/protocols/openmm.py", line 584, in _execute
    _compute_gradients(
  File "/data1/choderaj/wangl14/pydev/openff-evaluator/openff/evaluator/protocols/openmm.py", line 247, in _compute_gradients
    reverse_energies = _evaluate_energies(
                       ^^^^^^^^^^^^^^^^^^^
  File "/data1/choderaj/wangl14/pydev/openff-evaluator/openff/evaluator/protocols/openmm.py", line 122, in _evaluate_energies
    openmm_context = openmm.Context(system, integrator, platform)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data1/choderaj/wangl14/miniforge3/envs/water-models-experiments-cuda/lib/python3.12/site-packages/openmm/openmm.py", line 14269, in __init__
    _openmm.Context_swiginit(self, _openmm.new_Context(*args))
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
openmm.OpenMMException: Cannot create a Context for a System with no particles
```

 - Operating system

**Additional context**

The reason this is happening is because `create_openmm_system` requires at least vdW to add particles to a system. That means if you're trying to subset only on bonds, nothing will get added to the OpenMM system. Somewhat similarly, all the charge methods need to get added to the OpenMM system for a valid virtual site fit. I patched the valence issue [here](https://github.com/openforcefield/openff-evaluator/commit/2171630319dccd476f5a80824470b2d778d60dc0) and the nagl issue [here](https://github.com/openforcefield/openff-evaluator/commit/618b5127056430a68394699e9c7d4e37fba4e76c).

lilyminium:
## Description
Fixes #727 by adding vdW forces to the system_subset and a bunch of tests.

### DIFF
--- a/openff/evaluator/_tests/test_utils/test_openmm.py
+++ b/openff/evaluator/_tests/test_utils/test_openmm.py
@@ -735,7 +735,6 @@ def test_system_subset_improper_torsions_existing_force_field():
             "idivf1": 1.0,
         }
     )
-    force_field.register_parameter_handler(improper_handler)
 
     # Create a branched 4-atom topology
     topology: Topology = Molecule.from_mapped_smiles(
--- a/openff/evaluator/_tests/test_utils/test_openmm.py
+++ b/openff/evaluator/_tests/test_utils/test_openmm.py
@@ -607,7 +607,7 @@ def test_system_subset_constraints():
     )
 
     # Should have nonbonded force (for vdW) even though we're dealing with constraints
-    assert system.getNumForces() == 0 #?
+    assert system.getNumForces() == 0  # ?
     assert system.getNumParticles() == 2
     assert system.getNumConstraints() == 1
 
@@ -666,7 +666,10 @@ def test_system_subset_angles_with_full_force_field():
     assert system.getNumForces() == 1
     assert system.getNumParticles() == 3
     force_type = type(system.getForce(0)).__name__
-    assert force_type == "HarmonicAngleForce", f"Expected HarmonicAngleForce, but got {force_type}"
+    assert (
+        force_type == "HarmonicAngleForce"
+    ), f"Expected HarmonicAngleForce, but got {force_type}"
+
 
 def test_system_subset_proper_torsions_isolated_force_field():
     """Test that vdW handler is automatically included when dealing with ProperTorsions.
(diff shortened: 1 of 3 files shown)


openff/evaluator/_tests/test_utils/test_openmm.py
```diff
@@ -552,3 +552,238 @@ def test_extract_atom_indices():
...
+    constraint_handler.add_parameter(
+        {
+            "smirks": "[#1:1]-[#17:2]",
+            "distance": 1.0 * unit.angstrom,
+        }
+    )
+    force_field.register_parameter_handler(constraint_handler)
+
+    # Create a dummy topology
+    topology: Topology = Molecule.from_mapped_smiles("[Cl:1][H:2]").to_topology()
```
mattwthompson: (non-blocking) This could be a fixture

mattwthompson (approved):
Approved pending two blocking comments; I'm not asking to re-review but can if you think that would be valuable.

Please also
* update docstring/comments to the new implementation; IIUC since vdW handlers no longer need to be added when vdW parameters aren't being fit, they **shouldn't** be added or present in system subsets.
* add a one-liner in the release notes, doesn't need to be well-formatted

PR #734, opened 2026-02-11
Allow for empty topology labels for inapplicable keys

### LINKED ISSUE #728: Evaluator doesn't account for empty parameter labels when listing gradient keys to assign to a system
**Describe the bug**

To calculate gradients, Evaluator needs to work out which parameters are actually present in a particular property. It does so by checking if a property has a parameter type present when labelling molecules. However sometimes the parameter labels are empty when labelling molecules, for example in the case of LibraryCharges:

```python
ff = ForceField("openff-2.3.0.offxml")
mol = Molecule.from_smiles("N#N")
labels = ff.label_molecules(mol.to_topology())[0]
assert "LibraryCharges" in labels and not len(labels["LibraryCharges"])
```

**To Reproduce**

Apologies, will need an MRE for this too... 

**Output**

It triggers an IndexError once the workflow hits line 609 because the list is empty:

https://github.com/openforcefield/openff-evaluator/blob/884b8504dbb9d8922909b0d2f2629234dc8bc7fc/openff/evaluator/workflow/workflow.py#L590-L609

I think I recall the lipid co-optimizations running into this, too.

 - Operating system

**Additional context**

The [fix](https://github.com/openforcefield/openff-evaluator/commit/fd1a8f921529bbd8ceb3aa0c99f36743bcb97978) for this is fairly straightforward, at least.

lilyminium:
## Description
Fixes #728

### DIFF
--- a/openff/evaluator/_tests/test_workflow/test_workflow.py
+++ b/openff/evaluator/_tests/test_workflow/test_workflow.py
@@ -313,7 +313,9 @@ def test_find_relevant_gradient_keys(tmp_path):
         [
             *expected_gradient_keys,
             ParameterGradientKey(tag="vdW", smirks="[#6:1]", attribute="sigma"),
-            ParameterGradientKey(tag="LibraryCharges", smirks="[#5:1]", attribute="charge"),
+            ParameterGradientKey(
+                tag="LibraryCharges", smirks="[#5:1]", attribute="charge"
+            ),
         ],
     )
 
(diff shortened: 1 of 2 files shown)


openff/evaluator/_tests/test_workflow/test_workflow.py
```diff
@@ -287,6 +287,15 @@ def test_find_relevant_gradient_keys(tmp_path):
     )
     force_field.register_parameter_handler(vsite_handler)
 
+    # add constraints to test empty topology labels
```
mattwthompson: I assume this comment is out of date / accidentally erroneous? I don't really care what is used to test this, but library charges make sense as used here

mattwthompson (approved):
Approved pending me not being too wrong in my understanding of the problem (https://github.com/openforcefield/openff-evaluator/issues/728#issuecomment-3915849812)

Review log:
* Made sure source code change and test correspond to reported issue
* Grabbed the changeset locally to double check that the reported `IndexError` is triggered as expected

mattwthompson (approved):

PR #739, opened 2026-02-24
Forbid `ASH` residue name

### LINKED ISSUE #691: When a molecule is assigned ASH as a random name, it's incorrectly treated as ASP.
I think I ran into a hopefully one-off error generating a PDB using the Equilibration workflow where not all bonds were present, so just opening an issue to collect any potential future reports.

To reproduce using the same code-paths that trigger an error during PreequilibratedSimulation workflows:

```python
from openff.toolkit import Molecule, ForceField, Topology
import openmm
from openmm import app

unique_molecules = [
    Molecule.from_smiles("C1CCOCC1"),
    Molecule.from_smiles("Cc1ccccc1C"),
]

pdb_file = app.PDBFile("output.pdb")
topology = Topology.from_openmm(pdb_file.topology, unique_molecules=unique_molecules)

```

Error triggered:
```
File ~/micromamba/envs/openff-toolkit-release/lib/python3.12/site-packages/openff/toolkit/topology/topology.py:1517, in Topology.from_openmm(cls, openmm_topology, unique_molecules, positions)
   1506         if hill_formula in probably_missing_conect:
   1507             msg += (
   1508                 "This would be a very unusual molecule to try and parameterize, "
   1509                 "and it is likely that the data source it was read from does not "
   (...)
   1515                 'not specified in the standard residue connectivity table."'
   1516             )
-> 1517         raise ValueError(msg)
   1519 # The connected_component_subgraph function above may have scrambled the molecule order, so sort molecules
   1520 # by their first atom's topology index
   1521 topology_molecules_to_add.sort(key=lambda x: x[0])

ValueError: No match found for molecule C. This would be a very unusual molecule to try and parameterize, and it is likely that the data source it was read from does not contain connectivity information. If this molecule is coming from PDB, please ensure that the file contains CONECT records. The PDB format documentation (https://www.wwpdb.org/documentation/file-format-content/format33/sect10.html) states "CONECT records are mandatory for HET groups (excluding water) and for other bonds not specified in the standard residue connectivity table."
```

Looking at the bonds gives a clue -- `ASP` has none --

```python
import collections

counts = collections.defaultdict(set)

res = list(pdb_file.topology.residues())
for r in res:
    atoms = list(r.atoms())
    bonds = list(r.bonds())
    counts[r.name].add((len(atoms), len(bonds)))
counts
```
```
defaultdict(set, {'VLQ': {(16, 16)}, 'ASP': {(18, 0)}})
```

This was odd, sent me in a couple wrong directions for a few days. I think I've re-generated a fully-bonded PDB file now without changing anything in my environment, and am not super interested in trying to dig into what may have been a one-off error, but just noting it as a possible source of future issues.

[output.pdb.txt](https://github.com/user-attachments/files/21708143/output.pdb.txt)

mattwthompson:
## Description
Fixes #691

See [this comment](https://github.com/openforcefield/openff-evaluator/issues/691#issuecomment-3255106788) and more in the same issue for context

This is a quick and dirty hack that is not guaranteed to fix the surface-level problem[^1] and is guaranteed to not fix the underlying problem[^2]

## Todos
  - [x] Forbid `ASH` residue name

## Questions
- [ ] Is there any way to reliably[^1] test this? I think not

## Status
- [ ] Ready to go

[^1]: The failure case is basically a 1-in-a-large-number random occurrence and therefore I think impossible to reliably and reproducibly trigger
[^2]: A refactor of how intermediate Packmol PDB files are handled is necessary for this

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -8,6 +8,11 @@ Releases follow the ``major.minor.micro`` scheme recommended by
 * ``minor`` increments add features but do not break API compatibility
 * ``micro`` increments represent bugfix releases or improvements in documentation
 
+Current development
+-------------------
+
+* PR `#739 <https://github.com/openforcefield/openff-evaluator/pull/739>`_: Add ``ASH`` to list of forbidden residue names in Packmol wrapper
+
 0.5.2 - January 21, 2025
 ------------------------
 
--- a/openff/evaluator/_tests/test_utils/test_packmol.py
+++ b/openff/evaluator/_tests/test_utils/test_packmol.py
@@ -160,3 +160,24 @@ def test_amino_acids():
 
     for index, smiles in enumerate(smiles):
         assert trajectory.top.residue(index).name == amino_residues[smiles]
+
+
+def test_packmol_forbidden_residue_name_ash(monkeypatch):
+    import mdtraj
+
+    top = mdtraj.Topology()
+    chain = top.add_chain()
+    residue = top.add_residue("UNK", chain)
+    top.add_atom("C1", mdtraj.element.carbon, residue)
+    top.add_atom("C2", mdtraj.element.carbon, residue)
+
+    forced_choices = iter("ASHBCD")
+
+    def fake_choice(_):
+        return next(forced_choices)
+
+    monkeypatch.setattr(packmol.random, "choice", fake_choice)
+
+    packmol._generate_residue_name(residue, "CC")
+
+    assert residue.name == "BCD"
--- a/openff/evaluator/utils/packmol.py
+++ b/openff/evaluator/utils/packmol.py
@@ -207,6 +207,7 @@ def _generate_residue_name(residue, smiles):
         "NMA",
         "THY",
         "URA",
+        "ASH",
     ]
 
     amino_residue_mappings = {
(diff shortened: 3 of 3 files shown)


lilyminium (approved):
I think (and asked Claude to write) you could patch random.choice for the below test, otherwise LGTM, pending green CI. Thanks Matt!

```diff
diff --git a/openff/evaluator/_tests/test_utils/test_packmol.py b/openff/evaluator/_tests/test_utils/test_packmol.py
index 0c4b845..426f66f 100644
--- a/openff/evaluator/_tests/test_utils/test_packmol.py
+++ b/openff/evaluator/_tests/test_utils/test_packmol.py
@@ -160,3 +160,24 @@ def test_amino_acids():

     for index, smiles in enumerate(smiles):
         assert trajectory.top.residue(index).name == amino_residues[smiles]
+
+
+def test_packmol_forbidden_residue_name_ash(monkeypatch):
+    import mdtraj as md
+
+    top = md.Topology()
+    chain = top.add_chain()
+    residue = top.add_residue("UNK", chain)
+    top.add_atom("C1", md.element.carbon, residue)
+    top.add_atom("C2", md.element.carbon, residue)
+
+    forced_choices = iter("ASHBCD")
+
+    def fake_choice(_):
+        return next(forced_choices)
+
+    monkeypatch.setattr(packmol.random, "choice", fake_choice)
+
+    packmol._generate_residue_name(residue, "CC")
+
+    assert residue.name == "BCD"
```

PR #741, opened 2026-02-24
Add runtime limit to AWS instances

### LINKED ISSUE #740: Set max runtime for AWS runners
AWS runners sometimes get started but fail to be shut down, eg [this action](https://github.com/openforcefield/openff-evaluator/actions/runs/21603273240/job/62254492801). We should set a max runtime on our AWS runners to prevent this.

j-wags:
## Description
- [x] Fix #740 

## Status
- [x] Ready to go

### DIFF
--- a/.github/workflows/aws.yaml
+++ b/.github/workflows/aws.yaml
@@ -67,7 +67,7 @@ jobs:
         id-token: write
         contents: read
     needs:
-      - start-aws-runner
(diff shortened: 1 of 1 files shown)


mattwthompson (approved):
LGTM - the tradeoff (not having runaway jobs, but sometimes not having information about why a job was killed) is net positive to me

Thanks for the fix!

Issue #748, opened 2026-03-23
Facilities for easily manipulating storage


lilyminium:
**Is your feature request related to a problem? Please describe.**

Now that we use LocalFileStorage to store pre-equilibrated data, it would be convenient to easily modify (add, remove, subset, combine) storages for particular experiments. this is particularly relevant to the water models experiment because it would be convenient to have separate with-water and without-water file storages that we can easily combine for different experiments. A key part of this would be **copying data files** instead of **moving them** when an object is stored, as is currently the case for LocalFileStorage.

**Describe the solution you'd like**

I propose making a subclass called MutableFileStorage to handle these, to avoid unintended consequences on LocalFileStorage.

**Describe alternatives you've considered**

**Additional context**

mattwthompson:
I think we can close this with #749

PR #749, opened 2026-03-23
Add MutableLocalFileStorage subclass skeleton

lilyminium:
## Description
Related to #748  . This will function as a trunk PR -- I'll make additional PRs into this branch for easier review, then merge this if/when everything gets an approval.

As such no tests at the moment, but would be good to still get a quick review on this for anything that pops out.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - "main"
+      - "expand-lfs"
   schedule:
     - cron: "0 0 * * *"
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,6 @@ on:
   pull_request:
     branches:
       - "main"
-      - "expand-lfs"
   schedule:
     - cron: "0 0 * * *"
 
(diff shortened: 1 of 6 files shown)


mattwthompson (commented):
This seems on the right track, but yeah it's quite barebones. I certainly don't have objections or major suggestions for anything here since the bulk of the work comes in the follow-ups

lilyminium:
Ok -- I think that's each of the sub-PRs merged in! @mattwthompson did you want to re-review this, or is it good to go for `main`?

.github/workflows/ci.yaml
```diff
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - "main"
+      - "expand-lfs"
```
mattwthompson: I think we can drop this prior to merge

mattwthompson:
```suggestion
```

mattwthompson (approved):
I'm pleasantly surprised at the size of the final changeset - I think our approach of merging smaller changes into a staging trunk was correct (and I would do it the same if I could go back in time) but my concern that the final change would be illegibly large has not come true

I want to see what silicon has to say about it - but only for my curiosity, I don't think its feedback should block this being merged in

PR #751, opened 2026-03-23
Mlfs remove object

lilyminium:
## Description
Removes objects by storage keys. This may need to get expanded later (e.g. removing by substance or component...) but for now, storage key works.

Includes some copied code from other branches, should get reviewed after being rebased to reduce extraneous LOC.

### DIFF
--- a/openff/evaluator/_tests/test_storage/test_mutable.py
+++ b/openff/evaluator/_tests/test_storage/test_mutable.py
@@ -1,5 +1,6 @@
 import os
 import tempfile
+import uuid
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


openff/evaluator/_tests/test_storage/test_mutable.py
```diff
@@ -551,3 +552,134 @@ def test_subset_force_field_excluded_when_include_substance_active(
...
+        key = storage.store_object(obj, data_dir)
+
+        type_name = type(obj).__name__
+        assert key in storage._stored_object_keys[type_name]
+        storage.remove_object(key)
+        assert key not in storage._stored_object_keys.get(type_name, [])
+
+
+def test_remove_object_clears_hash_index():
+    """For HashableStoredData, the corresponding hash entry is removed."""
```
mattwthompson: Wrong class being described here? Though the test doesn't use this

mattwthompson (approved):
LGTM % one tiny error. (Non-blocking, no re-review)

lilyminium:
Thanks @mattwthompson !

PR #752, opened 2026-03-23
Combine storages

lilyminium:
## Description
Adds ability to combine/update storages. Rebasing/merging got complicated so now it's rebased off of #750 and should probably only get revewed after that's merged, to avoid repeating work ...


lilyminium:
Gah can I edit the destination?

PR #753, opened 2026-03-23
Mlfs subset storage

lilyminium:
Add abilities to filter storages based on including and excluding substances and components.


mattwthompson (approved):
LGTM, assuming this does what you need
 
Type annotations for the added methods would be nice (mostly because of the order I happened to read them in - the docstring in the public method is plenty) but non-blocking and no re-review

PR #754, opened 2026-03-23
Fix FilterBySmiles

lilyminium:
Fixes #722

I reproduced the failure in the first commit: https://github.com/openforcefield/openff-evaluator/actions/runs/23426363168/job/68142004946?pr=754

The issue was returning cls instead of data in the validator

PR #756, opened 2026-03-23
Fix ThermoML parsing

lilyminium:
Actually relates to this dimsim issue (https://github.com/openforcefield/dimsim/issues/53) but it's a shared issue as the ThermoML code was simply ported.

Edit: I messed up and rebased this off of #754 by accident, sorry


mattwthompson (approved):
This looks good but I found ~~two~~ one thing] I would like another look at; neither call for re-review if the outcomes are as expected

1. A blocking comment in the code about behavior when RDKit is not installed which can be resolved by you agreeing it's an okay and intentional change. No need to re-review just for that
2. ~~Triggering the failing test (https://github.com/openforcefield/openff-evaluator/pull/757), i.e. that check failing as expected~~ -- now resolved: test failing/skipped as expected: https://github.com/openforcefield/openff-evaluator/actions/runs/23498429535/job/68386169870?pr=757

PR #760, opened 2026-04-01
Add tautomer filter

lilyminium:
## Description

Adds functions for classifying, and filtering out, molecules that may be tautomers. The rules and examples are plotted below. There are a lot of considerations as to which and how much a form is dominant, so I don't recommend whitelisting some categories. Others, of course, are very common, like keto/enol tautomerism. 

I initially considered simply removing any molecule that had possible tautomers generaetd by RDKit's EnumerateTautomers function, but that would flag very simple molecules like acetone -- hence the need to overcomplicate and classify.

As mentioned very quickly in the FF fitting meeting today, this only would really apply to a small subset of ThermoML (~400 unique molecules), and the classified molecules largley did not raise red flags, save perhaps the ANNULAR_AZOLE category. This unfortunately was me throwing my hands up at the multitude of considerations of ring nitrogen tautomers and dumping them all into the one category. 

tautomer_rules

tautomers-summary

### DIFF
--- a/openff/evaluator/_tests/test_datasets/test_curation/test_filtering.py
+++ b/openff/evaluator/_tests/test_datasets/test_curation/test_filtering.py
@@ -38,10 +38,10 @@ from openff.evaluator.datasets.curation.components.filtering import (
     FilterByStereochemistrySchema,
     FilterBySubstances,
     FilterBySubstancesSchema,
-    FilterByTemperature,
-    FilterByTemperatureSchema,
     FilterByTautomers,
     FilterByTautomersSchema,
+    FilterByTemperature,
+    FilterByTemperatureSchema,
     FilterDuplicates,
     FilterDuplicatesSchema,
 )
@@ -1258,7 +1258,10 @@ def test_filter_by_tautomers_via_workflow():
     data_frame = pandas.DataFrame(
         [
             {"N Components": 1, "Component 1": "C"},
-            {"N Components": 1, "Component 1": "CC(=O)CC(=O)C"},  # beta-diketone, not in whitelist
+            {
+                "N Components": 1,
+                "Component 1": "CC(=O)CC(=O)C",
+            },  # beta-diketone, not in whitelist
         ]
     )
 
@@ -1285,9 +1288,7 @@ def test_filter_by_tautomers_no_match_removed_default():
 
 def test_filter_by_tautomers_whitelist_keto_enol_aliphatic_kept():
     """Acetone (KETO_ENOL_ALIPHATIC) must survive the default filter."""
-    data_frame = pandas.DataFrame(
-        [{"N Components": 1, "Component 1": "CC(C)=O"}]
-    )
+    data_frame = pandas.DataFrame([{"N Components": 1, "Component 1": "CC(C)=O"}])
 
     filtered_frame = FilterByTautomers.apply(data_frame, FilterByTautomersSchema())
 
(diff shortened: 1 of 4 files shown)


openff/evaluator/datasets/curation/components/filtering.py
```diff
@@ -1207,6 +1212,194 @@ def _apply(
...
+    of `categories_to_include` or `categories_to_exclude` to control which families
+    are allowed.
+
+    Notes
+    -----
+    **Suppression**
+
+    Category matching always uses suppression: if a molecule matches both a
+    parent category and a subordinate category, the subordinate is hidden and
+    only the parent is reported. The complete suppression table is defined in
```
mattwthompson:
```suggestion
    Category matching always uses suppression: if a molecule matches both a
    parent category and a subordinate category, its match with the subordinate is hidden and
    only its match with the parent is reported. The complete suppression table is defined in
```
I had trouble with the phrasing here; I get the behavior after reading through the tests, but didn't when first reading this. I could be wrong in how the schema is structured vs. how the filter behaves, but I think clarifying the behavior of molecules as passing through the filter vs. how different tautomeric states are structured in this class. In other words, it's not clear if the object of this sentence is the molecule or the tautomer definition(s)

openff/evaluator/_tests/test_datasets/test_curation/test_filtering.py
```diff
@@ -1212,3 +1214,224 @@ def test_curation_does_not_alter_precision():
...
+    with pytest.raises(ValidationError):
+        FilterByTautomersSchema(categories_to_include=None, categories_to_exclude=None)
+
+    with pytest.raises(ValidationError):
+        FilterByTautomersSchema(
+            categories_to_include=["AMIDE_IMIDIC_ACID"],
+            categories_to_exclude=["BETA_DIKETONE"],
+        )
+
+    with pytest.raises(ValidationError):
```
mattwthompson, line 1233: Please add `match=...` to each of these, since Pydantic raises the same error for each failure case

mattwthompson (commented):
I'm still working through this, but I'm leaving a partial review in case my browser is hungry enough to eat these when I come back to it later

openff/evaluator/datasets/curation/components/tautomers.py
```diff
@@ -0,0 +1,242 @@
...
+    seen = set()
+    # Enumerate always includes the input molecule, so no need to pre-seed `seen`.
+    for t in _TAUTOMER_ENUMERATOR.Enumerate(rdmol):
+        smi = Chem.MolToSmiles(t, isomericSmiles=True)
+        if smi not in seen:
+            seen.add(smi)
+            try:
+                tautomers.append(Molecule.from_smiles(smi, allow_undefined_stereo=True))
+            except Exception:
+                pass
```
mattwthompson: What failures are lost in this? The there cases in which `Chem.MolToSmiles` returns something but `Molecule.from_smiles(smi, allow_undefined_stereo=True)` chokes are

* Multi-component (hopefully not triggered by now!)
* Radicals (ditto)
* Kekulization failures ... ?

mattwthompson (approved):
Approving (comments are non-blocking, re-review only if you wish) leaning somewhat on this being functional in your live use cases. This is maybe 90% of the way towards a thorough review in ~half the time, a tradeoff I'm happy with given that this codebase is only somewhat public-facing and the key stakeholder appears happy with the behavior.

I didn't go through every SMARTS pattern in the rule definitions in the source code - I can if we want to - but I did look fairly closely at the tests and the behavior. This took me a few passes through to understand, including refreshing myself on organic chemistry that's been lost to memory. (I'm not suggesting process improvements, the changes are highly interconnected it's not clear how this could be broken up; most of the work is setting up the logic, so separate PRs for adding in tautomerization chemistries bit-by-bit would be ... perhaps tedious.)

openff/evaluator/datasets/curation/components/tautomers.py
lilyminium: I think this was some preemptive caution since I have seen RDKit molecules fail to round trip before due to some issue with the bond table. We could probably undo the try/except -- I tested the public ThermoML I had cached and it didn't seem to run into any issues in the end.

mattwthompson: Either approach is fine by me - in principle undoing it would be better but it may not be worth the uncertainty

PR #783, opened 2026-07-24
Add soft propertytype filter

### LINKED ISSUE #738: A filter that grabs all dhmix data, then selects other (e.g. density) data with the same compounds, then fills out more density data as needed
**Is your feature request related to a problem? Please describe.**

ThermoML has way more density than dhmix data. We currently get the proportions roughly equal by requiring that mixtures must have both density and dhmix data points at a very early stage in filtering, but this eliminates quite a lot of dhmix data that would be valuable in fits. We need some kind of filter that keeps all dhmix data but can select the densities down to roughly equal proportions. The method in the title is one way.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

lilyminium:
## Description
Fixes #738 

Adds `FilterByCoreAndAdditionalPropertyTypes`, which is a modification of FilterByPropertyTypes but "softer". FilterByPropertyTypes requires that a substance has to be present in all property types to be kept, which was stripping out a lot of dhmix data that had *similar* substances present with the densities, but not always the *same* (e.g. CCC vs CCCC) . This filter instead lets you keep all the properties in "core" (e.g. keep all the dhmix data) and then select from the "additional" dataset (in this example, densities) for similar substances.

### DIFF
--- a/openff/evaluator/_tests/test_datasets/test_curation/test_filtering.py
+++ b/openff/evaluator/_tests/test_datasets/test_curation/test_filtering.py
@@ -1225,15 +1225,18 @@ def _substances_for_property(data_frame: pandas.DataFrame, property_type: str) -
 
 class TestFilterByCoreAndAdditionalPropertyTypes:
 
+    property_types = ["Density", "EnthalpyOfMixing"]
+
     def test_validate_filter_by_core_and_additional(self):
+        """Generally test valid arguments for schema and set expectations"""
         # Valid schema
         FilterByCoreAndAdditionalPropertyTypesSchema(
             core_property_types={"Density": None},
-            additional_property_types={"SurfaceTension": None},
+            additional_property_types={"EnthalpyOfMixing": None},
         )
         FilterByCoreAndAdditionalPropertyTypesSchema(
             core_property_types={"Density": [1]},
-            additional_property_types={"SurfaceTension": None},
+            additional_property_types={"EnthalpyOfMixing": None},
             scale_factor=0.0,
         )
 
@@ -1248,7 +1251,7 @@ class TestFilterByCoreAndAdditionalPropertyTypes:
         with pytest.raises(ValidationError):
             FilterByCoreAndAdditionalPropertyTypesSchema(
                 core_property_types={"Density": None},
-                additional_property_types={"SurfaceTension": None},
+                additional_property_types={"EnthalpyOfMixing": None},
                 scale_factor=-1.0,
             )
 
@@ -1256,7 +1259,7 @@ class TestFilterByCoreAndAdditionalPropertyTypes:
         with pytest.raises(ValidationError):
             FilterByCoreAndAdditionalPropertyTypesSchema(
                 core_property_types={},
-                additional_property_types={"SurfaceTension": None},
+                additional_property_types={"EnthalpyOfMixing": None},
             )
         with pytest.raises(ValidationError):
             FilterByCoreAndAdditionalPropertyTypesSchema(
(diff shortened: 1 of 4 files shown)


openff/evaluator/datasets/curation/components/filtering.py
```diff
@@ -42,6 +44,9 @@
 
 ComponentEnvironments = List[List[ChemicalEnvironment]]
 MoleFractionRange = Tuple[confloat(ge=0.0, le=1.0), confloat(ge=0.0, le=1.0)]
+# A mapping of property type -> optional list of allowed component counts
+# (``None`` — or an empty list — means no ``n_components`` restriction).
+PropertyTypeFilter = Dict[constr(min_length=1), Optional[List[PositiveInt]]]
```
mattwthompson, line 52:
```suggestion
# A mapping of property type -> optional list of allowed component counts
# (an empty list means no ``n_components`` restriction).
PropertyTypeFilter = Dict[constr(min_length=1), List[PositiveInt]]
```

This should make the annotations simpler and IIUC behave the same

openff/evaluator/datasets/curation/components/filtering.py
```diff
@@ -1212,6 +1217,420 @@ def _apply(
...
+        "int(scale_factor * core_count) substances are targeted. The overlap — "
+        "additional-type data already present for core substances — is always "
+        "retained. Set to 0 for overlap-only (no gap-fill).",
+    )
+    n_components: Optional[List[PositiveInt]] = Field(
+        None,
+        description="If set, only substances with this many components qualify "
+        "for this additional type. An empty list is treated the same as None: "
+        "no n_components restriction.",
+    )
```
mattwthompson, line 1256: Struggling to think of how this would be longer than 1 / the tests only use 1-length lists

openff/evaluator/datasets/curation/components/filtering.py
```diff
@@ -1212,6 +1217,420 @@ def _apply(
...
+
+    def _grow_diversity_reference(self, substance):
+        """Fold a freshly picked substance's fingerprints into the MaxMin reference."""
+        for smi in substance:
+            if smi not in self._ref_smiles:
+                self._ref_smiles.add(smi)
+                fp = _morgan_fp(smi)
+                if fp is not None:
+                    self._ref_fps.append(fp)
+
```
mattwthompson, line 1454: I infer that it doesn't matter these reference attributes (I think `set`s) are un-ordered and could be different length?

openff/evaluator/datasets/curation/components/filtering.py
mattwthompson: Now I gather that this is the case because you're focused on only one additional type?

mattwthompson (changes requested):
I'm partway through reading this (and haven't found anything objectionable) but I'm struggling to understand the complexity of the underlying logic in the filters. Could you add type annotations to method signatures in `_GapFiller`? We don't run a type-checker here so I don't care about them being strictly correct, but they would help me follow how different variables are passed through

openff/evaluator/datasets/curation/components/filtering.py
lilyminium: Ah no -- that's more becuase I was being lazy. The use case I wrote this for adds Density as the AdditionalPropertyType with n_components=[1, 2] because we're looking for both pure and mixture densities, but not (e.g.) mixtures with 3 components. would you want a test of this?

openff/evaluator/datasets/curation/components/filtering.py
lilyminium: No -- could probably be a set too, but I think fingerprints aren't hashable.

(edit for clarity) -- as in, your inference here is correct.

openff/evaluator/datasets/curation/components/filtering.py
lilyminium: oh oops that did break something -- I might revert for now

lilyminium:
Closing and reopening to jumpstart CI, which I think I might have broken ...

openff/evaluator/datasets/curation/components/filtering.py
lilyminium: I added a test for this.

mattwthompson: This makes sense now - ternary mixtures aren't a focus of these tests, and probably shouldn't be

openff/evaluator/_tests/test_datasets/test_curation/test_filtering.py
```diff
@@ -1723,6 +1723,43 @@ def test_filter_by_core_and_additional_empty_n_components(self):
...
+        density_subs = _substances_for_property(filtered, "Density")
+        # Core overlap: binaries that also have density
+        assert ("CC", "O") in density_subs
+        assert ("CCC", "O") in density_subs
+        # Gap-filled substances
+        assert ("CC",) in density_subs
+        assert ("CCCC", "O") in density_subs
+        # excluded
+        assert ("CC", "CCC", "O") not in density_subs
+        assert ("CF", "N") not in density_subs
```
mattwthompson, line 1761: Thanks, this was helpful for me

mattwthompson (approved):
Thanks for iterating, with a second pass through it's come together for me

mattwthompson:
Not sure what was going on with the automation, strong suspicion of an unreported GitHub outage

lilyminium:
Thanks for the review Matt!

PR #786, opened 2026-08-13
Discretize query logic

lilyminium:
## Description
I ran into an issue where I uploaded boxes into the LocalFileStorage with representative mole fractions, but the current logic of matching the substance mole fraction exactly didn't work with this (especially in cases where the mol count was 999). (a representative MWE of the issue is added as a test). This PR makes the check a bit fuzzier and discretizes the molecule counts.


mattwthompson (approved):
LGTM, tests fail without the patch and the change is a reasonable improvement

